PR #10, opened 2019-11-18
Create dipstaging document

RussellMcOrmond:
Filled in descriptions of some of the existing fields.  If we want to use https://json-schema.org/ at this stage, I have much more reading to do.

### DIFF
--- a/Databases/dipstaging.md
+++ b/Databases/dipstaging.md
@@ -1,0 +1,17 @@
+
+## _update/basic fields
+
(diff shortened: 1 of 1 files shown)


SaschaAdler:
You can see [an example](https://github.com/crkn-rcdr/parliament-metadata/blob/master/schemas/attachment.json) of a JSON Schema file that we've used in our (currently private) Parliament repo.

Issue #11, opened 2019-11-26
The 'tx' field (Internalmeta, cosearch)


RussellMcOrmond:
Currently the 'tx' field is filled in by Hammer by removing all XML tags (and thus positional information) from the OCR XML.  This is stored within hammer.json (array indexes above 0), and then stored in cosearch documents in a 'tx' field.

In 'cosearch' the ID's are slug based, so even canvas documents will need to be generated per-manifest.  This suggests that this field should be generated based on something 'manifest' related, not 'canvas'.

Press manages the cosearch documents, and instead of only getting the "_rev" from the previous document it can also grab previously stored data including a previously stored 'tx' field.  Only if the 'tx' field is missing, or needed to be updated, would there be a need to process the OCR XML data.

We haven't yet decided how/when OCR data will be loaded, but it won't be via Smelter as OCR data currently stored in legacy platform AIPs are being considered access data.  However it is loaded there will be a need to set a field (a date) which causes processing.

Possible new fields in ''canvas' documents:

* ocrxmldate - date when the OCR XML was last updated

Whatever tools update OCR data would also trigger the update of "updatereq" of any manifests which include this canvas.  We would need to have a _view to obtain this list.  Press, after pulling up the 'canvas' documents referenced within the 'manifest' document would also be able to detect if OCR xml data had been updated.

RussellMcOrmond:
Relating to triggering databases....

Is the intention with administrative tools that there would be a back-end server running, or that the web client would be doing all the work?  If we intend to put OCR data within a zip file, we will need to be able to extract that OCR data from the zip.  Is this something that will be done on the server or the client?

SaschaAdler:
Is anything stopping Smelter from loading OCR data stored in the legacy preservation system as a one-time thing? Future OCR updates would be handled by a front-end tool (as well as some back-end utility, maybe, if unzipping and figuring out where the OCR data should go proves difficult).

We do need to figure out the exact mechanism for updating OCR once we aren't loading it from the preservation platform. I assume we won't be updating OCR image-by-image, and we're going to have to make it clear that the access manifests that will be updated aren't the same thing as preservation packages.

RussellMcOrmond:
Code simplicity is the only reason we might want to avoid that.  Keep Smelter more easily testable.

If Smelter stores the METS sequence in dipstaging, along with all the canvas noids that were allocated for the members of those sequences, then a separate tool can load the OCR data that first time.

[...]

PR #12, opened 2019-11-26
Starting 'collection', 'manifest', 'canvas'

RussellMcOrmond:
Documentation in simple markdown.  Wasn't sure how to do the collaboration we need at this point using the JSON schema, which we can firm up later.

This is a first pass at documenting what a 3-way split of the data in 'internalmeta' would look like, with the data in hammer.json no longer being an attachment and using noid's for document IDs.

Once we settle on what to do with fields we'll need to split what is in the design document https://github.com/crkn-rcdr/cihm-metadatabus/tree/master/couchdb/internalmeta/design/tdr  into what would be correct for the 3 separated databases.

Some questions:

1. What to do about 'tx'.  #11
1. 'parent' field within issues of a series.  Should this be a noid looked up by Smelter and staff tools, or should this be 'slug' related and looked up by press?
1. Do we want to document more of the dmd.* fields, providing a complete list that might exist?  We didn't in the past, and only discussed specific fields in the cosearch-copresentation spreadsheet.
1. For the potentially deprecated fields, do we want to explore which ones used by CAP (Solr searches and/or copresentation), or just generate any still listed in [cosearch-copresentation](https://crkn.sharepoint.com/:x:/s/platform/Ed9hL3v-q_xMr9XGgMAd9s0BB0LrkwZjc8Ao_9IE4FkBXg?e=Pv2COM) and worry about cleaning up later?


RussellMcOrmond:
* In manifest.schema.json the labels are missing (label of manifest, label of each canvas).  Labels are a manifest thing, not a canvas thing, as the same canvas in multiple manifests would't necessarily have the same labels.
* We should decide how we want the link between the access and preservation platforms to be stored.  Do we store the canvas noids within dipstaging?  We may want this to be visible on the access platform at some point (this canvas stored within this AIP id, and with Archivematica the following persistent ID).
* what is 'deposits' in dipstaging?  It exists in the _update document, but does it have meaning for us?

SaschaAdler:
I've added the labels in the most recent commit.

What kind of link between preservation and access do you think we'll need? We're going to need to add preservation image/PDF URLs to this schema, until we start creating access copies. Do we need to store the manifest/canvas NOIDs of the new access objects in preservation also? If so, dipstaging seems like a reasonable place to do it.

`deposits` in dipstaging was meant to be the queue/log of all Smelter operations that had taken place on a given DIP(/AIP). Since we're now envisioning Smelter as a one-time thing, I'm not sure if we need to keep a complicated structure for this, instead of using some kind of `smeltedOn` timestamp.

SaschaAdler:
I've added the master/download info into the manifest and canvas schemas.

How badly are we going to need to directly reference dmd/OCR files in Swift? Since there is a scheme for where these files will be located, tools could theoretically use the NOID plus a few Swift lookups to find every possible file they could need. On the other hand, since there are multiple possible formats, it might make sense for us to list either the URLs for the files directly, or flag which formats we're using, which might looks like `"dmd":"issueinfo"` or `"ocr":"txtmap"`.

SaschaAdler:
I'm about to delete the Markdown files you've added; I'll reproduce them in this comment.

---

# dipstaging

## _update/basic fields

Some fields are automatically filled in when the tool 'reposync' (Part of [CIHM::META](https://github.com/crkn-rcdr/CIHM-Meta) ) posts data to dipstaging's [_update/basic](https://github.com/crkn-rcdr/cihm-metadatabus/blob/master/couchdb/dipstaging/design/sync/updates/basic.js).

* deposits - empty array created when document first created
* repos - array of repositories which has this AIP
* reposManifestDate - date on the AIP manifest.  Used to detect when the AIP has been updated, so this date should be stored with any processing.
* reposDate - the date when the above two field were filled in.
* METS - array of hashes indicating the 'path' and 'md5' of every metadata.xml file within the AIP
* METSManifestDate - date on the AIP manifest when this field was filled in
* METSDate - the date when the above two field were filled in.
* updated - date of when this document was last updated

---

# collection documents

One of the sources of data are similar to what was previously in 'internalmeta' for "series" documents, including data that was in hammer.json array index 0.
Notes in [manifest.md](manifest.md) apply to those documents filled in from METS records, which might be a one-time operation for this database.


## fields to be stored within couch documents

* dmd - hash of fields extracted from stored descriptive metadata, previously components stored as array index 0 of the hammer.json attachment. These are extracted from a temporarily generated "CMR" XML using [hammer2co.xsl](https://github.com/crkn-rcdr/CIHM-METS-parse/blob/master/lib/CIHM/METS/resource/xsl/hammer2co.xsl).  Common fields include:
  * identifier - array if identifiers
  * label - label from descriptive metadata
  * lang - array of languages
  * no - array of notes
  * no_source - array of notes indicating source
  * pubmin - minimum date in date range (first date for DC, MARC 260c)
  * pubmax - maximum date in date range (second date for DC, MARC 260c)
* collectionDate - date when the list of collections was modified (used by Press)
* collections - array of collections
* created - date document was created
* hammer - hash of fields used by Hammer to know when to run, and to report on success/failure
  * message - message from Hammer (die/warn)
  * dmddate - date of processed descriptive metadata
  * status - boolean for success/failure of Hammer (Did it "die")
  * date when hammer ran
* label - label from METS, most often extracted from descriptive metadata when METS record created
* approved - date when collection made public (unset means staff only)?
* press - hash of fields used by Press to know when to run, and report on success/failure
  * status - boolean for success/failure
  * messsage - die/warn messages
  * date - date when Press run, used in comparing other dates to know when to next run
* slug - string representing slug
* updated - date of last update
* updatereq - date when a press update was requested (used as part of press handling of re-processing parent records)

---
# manifest documents

Database fields are similar to what was previoiusly in 'internalmeta', including data that was in hammer.json array index 0


## fields to be stored within couch documents

* dmd - hash of fields extracted from stored descriptive metadata, previously components stored as array index 0 of the hammer.json attachment. These are extracted from a temporarily generated "CMR" XML using [hammer2co.xsl](https://github.com/crkn-rcdr/CIHM-METS-parse/blob/master/lib/CIHM/METS/resource/xsl/hammer2co.xsl).  Common fields include: 
  * identifier - array if identifiers
  * label - label from descriptive metadata
  * lang - array of languages
  * no - array of notes
  * no_source - array of notes indicating source
  * pubmin - minimum date in date range
  * pubmax - maximum date in date range
  * seq - sequence stored within issueinfo
  * pkey - ID of the series within issueinfo
* canonicalDownload - reference to a download (usually PDF)
* canonicalDownloadMime - MIME-type of download
* canonicalDownloadSize - size (bytes) of download
* collectionDate - date when the list of collections was modified (used by Press)
* canvases - array of hashes
  * canvasID - noid of canvas
  * label - string representing the label
* collections - array of collections
* created - date document was created
* hammer - hash of fields used by Hammer to know when to run, and to report on success/failure
  * message - message from Hammer (die/warn)
  * dmddate - date of processed descriptive metadata
  * status - boolean for success/failure of Hammer (Did it "die")
  * date when hammer ran
* label - label from METS, most often extracted from descriptive metadata when METS record created
* approved - date when manifest (and its canvases) were made public (unset means staff only)?
* parent? - reference to parent for issues. Should point to noid of collection, determined by Smelter?
* press - hash of fields used by Press to know when to run, and report on success/failure
  * status - boolean for success/failure
  * messsage - die/warn messages
  * date - date when Press run, used in comparing other dates to know when to next run
* slug - string representing slug
* updated - date of last update
* updatereq - date when a press update was requested (used as part of press handling of re-processing parent records, so this might be moved to 'collection' database.

## Fields that should be deprecated and/or stored/generated elsewhere

* dmd.depositor - currently based on the preservation ID (part before the first '.').  We should store a reference back to the preservation ID and generate this as current cosearch makes use of this.
* dmd.key - becomes slug
* dmd.pkey - refers to parent for an issue of a series, which we will need to think about.
* dmd.type - always "document"
* METS - hash of references to METS records, now in 'dipstaging' database
* METSDate - in 'dipstaging'
* METSManifestDate - in 'dipstaging'
* repos -- in 'dipstaging', never used within cosearch/copresentation
* reposDate - in 'dipstaging'
* reposManifestDate - in 'dipstaging'
* attachInfo - no attachments stored
* hammerfields - previously a list of fields from hammer within hammer.json attachment.  Fields now listed under dmd hash.
* pageinfo - relevant diagnostic information within 'canvas' database
* pubmin - duplication of dmd.pubmin
* seq - duplication of dmd.seq
* type - was always 'aip'
* sub-type - was one of "series" or "document", and series records now moving to 'collection' database


If we use any of these fields within cosearch(Solr) or copresentation then they can be generated by Press.

---

# canvas data

Database fields are what was previously in 'internalmeta' in the array within hammer.json index above 0.

## Fields to be stored within couch documents

* canonicalDownload - reference to download, usually a PDF file
* canonicalDownloadMime - MIME type of download
* canonicalDownloadSize - size (bytes) of download
* canonicalMaster - reference to master image
* canonicalMasterHeight - heigh of image
* canonicalMasterMime - MIME type of image
* canonicalMasterSize - size (bytes) of image
* canonicalMasterWidth - width of image
* tx - text-only string extracted from OCR XML

## Fields that should be deprecated and/or stored/generated elsewhere

* pkey - multiple manifests could be parent.
* label - labels will be in the array of canvases within manifests
* type - since databases are separate, the type="page" from METS is redundant as all canvases are of that type in METS.
* seq - sequence becomes the index within the manifest, so shouldn't be stored here.
* key - This is the manifest slug plus index within manifest.
* depositor - this is a component of the slug of the manifest
* identifier - array of identifiers (slugs of manifests, without depositor, with a sequence index appended)

If we use any of these fields within cosearch(Solr) or copresentation then they can be generated by Press.

PR #15, opened 2020-02-13
Default HAProxy config

SaschaAdler:
Yes, I moved stuff around in it for no good reason.

This uses IP addresses instead of DNS, so that switching what your local DNS points to on the regular doesn't require restarting everything.

Access platform repos are going to have default configs that expect this is running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+haproxy/*.crt
+haproxy/*.key
+haproxy/*.pem
+haproxy/haproxy.cfg
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "3.5"
+
+services:
+  haproxy:
+    image: haproxy:2.1.2-alpine
+    container_name: haproxy
+    ports:
+      - "80:80"
+      - "443:443"
+    volumes:
+      - ./haproxy:/usr/local/etc/haproxy:ro
+    networks:
+      access-platform:
+        ipv4_address: 172.30.0.2
+networks:
+  access-platform:
+    name: access-platform
+    driver: bridge
+    ipam:
+      config:
+        - subnet: 172.30.0.0/16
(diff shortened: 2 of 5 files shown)


RussellMcOrmond:
I think you moved things around for a good reason -- grouping together makes the fact these things aren't strictly sequential more clear.

I assume this is all tested at your end?  If so, then looks great to me.

Is there a desire to add notes on how to generate logging for debug purposes?

https://www.haproxy.com/blog/introduction-to-haproxy-logging/

> If you are running HAProxy inside of a Docker container and you’re using HAProxy version 1.9, then instead of sending log output to a Syslog server you can send it to stdout and/or stderr. Set the address to stdout or stderr, respectively. In that case, it’s also preferable to set the format of the message to raw, like so:
> 
```
global
  log stdout  format raw  local0  info
```

SaschaAdler:
I did want to figure out debug logging, and so I'll give that a go before I merge.

RussellMcOrmond:
In the puppet config I create a /etc/canadiana/haproxy-test which is a quick way to do a config file test.  I run it before sending a HUP to the container, allowing haproxy to load new config without any current requests being dropped.  Quite nice.

The test is far more thorough than what Apache does...

```
 # service puppet restart
 # /etc/canadiana/haproxy-test
 # docker kill -s HUP haproxy
```

```
root@mamirolle:~# cat /etc/canadiana/haproxy-test
#!/bin/sh

docker run -it --rm --name haproxy-syntax-check -v '/etc/canadiana/haproxy/:/usr/local/etc/haproxy/:ro' haproxy:2.1.2-alpine -c -f /usr/local/etc/haproxy/haproxy.cfg

root@mamirolle:~# 

```

SaschaAdler:
I did run that on this config at some point, although I've altered it since. It's a nice thing to test before committing updates.

PR #18, opened 2020-02-14
Some minor updates for logging and redirects.

RussellMcOrmond:
The logging doesn't otherwise indicate http/https, or what host was being requested.
An extra '/' was being added to the URLs being redirected.


SaschaAdler:
Looks good.

Issue #25, opened 2020-04-08
Move design documents from cihm-metadatabus


RussellMcOrmond:
https://github.com/crkn-rcdr/cihm-metadatabus/tree/master/couchdb currently has design documents.  Idea is to move these to https://github.com/crkn-rcdr/Access-Platform/tree/master/Databases/design-documents (sub-directories as supported by https://github.com/crkn-rcdr/kivik )

PR #26, opened 2020-04-08
Documents copied from cihm-metadatabus

RussellMcOrmond:
This solves #25


RussellMcOrmond:
It might be appropriate to clean this up to a common style now, or do that later.  I'll let the dev team decide (and possibly help with suggestions/etc).

SaschaAdler:
I'm just looking into formalizing how we use prettier for Couch 1 design docs. I've run prettier on these before, but a newer version is inserting a space between `function` and (e.g.) `(doc)` on all of the anonymous function definitions we're using here. I can't imagine Couch would have a hard time parsing that but it might be worth testing.

RussellMcOrmond:
I think we should merge, and then consider that a separate ticket to make the changes in a branch and test?

PR #36, opened 2020-04-21
Updates to CouchDB design documents for new access platform

RussellMcOrmond:
Some new views are needed for new tools that use the new databases.

This might also be a good time to create consistency in style.


SaschaAdler:
The only stylistic issue I see is how everything but the top line is indented an extra two spaces in aliases.js.

The Prettier extension can handle that without problems. You'll notice if using it (as of Prettier 2.0) that they made a decision to have anonymous functions using the `function` keyword have a space between `function` and the function's arguments, e.g. `function(doc)` becomes `function (doc)`. I agree with their motivation for doing this (it looks like `function` is the name of the function, otherwise), but this change would affect all of the already Prettier-ed design docs we have.

RussellMcOrmond:
Seems I forgot to restart after adding the Prettier plug-in.  I've done so and upgraded the version of VSCode to a more recent update.

That changes the default formatting so that the double rather than 4 spaces is the default.

I don't see it making notes about the lack of space after the function -- is it on your end?


I note the suggestion to convert to ES6 module, which we can't do yet.

SaschaAdler:
The changes look good.

The ES6 module notice isn't coming from Prettier, but VSCode itself. I don't actually think we can do anything about it. Theoretically we can edit the module wrapper around the function, since only the function ends up being sent to Couch, but I don't think it's a priority.

RussellMcOrmond:
You had some comments earlier about doc['_id'] vs doc.aliases.  Did we want to do anything specific with that, for any of these or other views?

I'll be leaving this branch open for a bit as I work though a few tickets relating to SPLIT.4 .  I've deployed the current design documents to the production database servers.

SaschaAdler:
I think `doc._id` is actually legal, but maybe it isn't in Couch's old parser. Ideally we use dot-notation for every key where it works (i.e. not something like `doc["hyphenated-key"]`).

We'll probably want to sweep through every design doc to sort that out, eventually. No rush on that, though.

RussellMcOrmond:
The rebase was noisier than I thought it would be.  Interesting..

RussellMcOrmond:
Seems appropriate to review and merge, creating new PRs for other functionality.

PR #48, opened 2020-08-06
Moving metadata bus specific views and updates to separate design doc…

### LINKED ISSUE #41: Move manifest/_design/access/_view/canvasnoids into a separate design document
I've had to edit the other views in the manifest/_design/access ddoc, and I'm twiddling my thumbs waiting for the indexer to get through every manifest's canvas before I'm able to see the results.

Not a big deal, but if I need to make more edits to these views, it might be nice to not have it trigger a rebuild of the canvas views.

RussellMcOrmond:
Would resolve #41 , and start cleaner with Hammer2 work.


RussellMcOrmond:
The views in "access" calculated so much faster this time compared to "metadatabus" (which is still calculating on my desktop).  I can see many ways this is an advantage, and suggests the O(n) database pull isn't as expensive as I was thinking it was.

SaschaAdler:
Looks good to me.

RussellMcOrmond:
```
-rw-r--r-- 1 couchdb couchdb 22612025507 Aug  6 02:11 11de5555074fa45072c1652b32890a4f.view
-rw-r--r-- 1 couchdb couchdb  1099419790 Aug  6 01:20 97fb5d19b4d910012cd6b492d17eecd3.view
-rw-r--r-- 1 couchdb couchdb 23454335203 Aug  5 19:00 d7d93fd0da496d84a3c17da01bba0a5c.view

```

A note about views size.  The views calculation started at 01:05:58, and the smaller "access" design doc was done 15 minutes later, but the "metadatabus" design doc took over an hour.

The larger/older file is the combined one, so a little less than the combined size of the two individual files.

PR #55, opened 2020-09-24
Descriptive Metadata loader: Database schemas (dmdtask), port to Access Admin interface, etc

RussellMcOrmond:
This is a sibling to https://github.com/crkn-rcdr/sapindale/pull/91 . part of SPLIT.12


RussellMcOrmond:
In discussion with @SaschaAdler it was suggested that there be a status field to indicate what stage the data was in.  We can add this if it is seen as necessary, although the existence or not of specific fields or values indicates the status.

* Only an _id indicates a new document
* When split.requestDate is set, "mdname" and "mdtype" must have been set and "mdname" referencing an attachment that exists. "depositor" is optional.
* When split.succeeded is true, items[] will have array values.
* When store.succeeded is set, at least one items[].validated must be true with an item[].copyto set, otherwise it will only generate an error.

While I see the need for documentation of the fields, I'm not sure if strict validation of stages within the schema is the best place.  The logic already needs to exist in other code (in Lapin and Sapindale).  Is there a good place to put this logic documentation, separate from specific implementations?

RussellMcOrmond:
@SaschaAdler 

This is ready to be reviewed and merged at your convenience.

The change that didn't make it in previously relates to a .json attachment that is the output of CIHM::Meta::dmd::flatten .  It seemed appropriate for staff to be able to click on a dmd record from a multi-record upload, and be able to see both the XML and how we will interpret the XML.  

pubmin and pubmax are also in the items[] array, as the two fields that should probably be highlighted in a table after uploading (as part of the review).

RussellMcOrmond:
I have created `dmdtask` on Iris and Gouda in Toronto, and set up replication between the couch servers. This is what we have been doing for the other new-Access platform databases.

Sapindale and the demos are set up on Iris https://github.com/crkn-rcdr/puppet/blob/master/manifests/nodes/iris.pp#L681

As is Lapin, pointing at Iris which is what the new Sapindale will point to when deployed.
https://github.com/crkn-rcdr/puppet/blob/master/manifests/nodes/iris.pp#L756

I didn't deploy the design documents, and thought it best for Sascha to review and merge.

SaschaAdler:
Quick question for the schema: are `mdname` and `mdtype` ever going to be different?

I might quibble about the names of some of these fields but otherwise things look fine, although it would maybe be nice to know if the properties of each item are required or not.

(Here's my quibble: if we're already in the `dmd` namespace of sorts, we probably don't need to preface `mdname` and `mdtype` with `md`. Also, I'm a fan of using camel case for longer names with multiple words in them.)

SaschaAdler:
I think I'd also like to change `depositor` to `prefix`, as per earlier discussions about how we might be trying not to encourage the concept of "depositor" in the Access Platform. I also think the prefix should include the period, so that we can also handle situations like `prefix === "oocihm.lac_reel"`.

RussellMcOrmond:
> Quick question for the schema: are mdname and mdtype ever going to be different?

So, the schema has a typo for `mdname` in that it isn't supposed to have the enum.  The name that was put there is the name given to it by the person uploading.

This can be changed to ignore the name of the file someone sends and always assume the attachment is a specific hard-coded name. If we wanted to have an appropriate file extension, it could be determined based on the mdtype ('.csv' and '.mrc').  If you wanted to get rid of that field entirely you could update the  src/components/DMD/Editor.svelte in Sapindale to match.

Also agreed on `prefix` rather than `depositor`, with the change also needing to be reflected in Sapindale.

Same with the `md*` thing -- just need to make Sapindale match, and I'll update the Perl side of things to match the updated schema once I'm on that task.

RussellMcOrmond:
Some notes for @DivyaKrishnan26 or @brittnylapierre , depending on who is later assigned this work.

When the existing Sapindale demo was updated to not talk to CouchDB directly, I felt it appropriate to simplify the attachment.  If you look at  https://github.com/crkn-rcdr/sapindale/blob/main/src/routes/dmd.json.js , you will notice that the attachment is now always named "attachment". 


```
{
  "_id": "e413481726c43d81b0dfc33d9e113941",
  "_rev": "5-749a6334eb0e95a77f1517dcc1166d31",
  "updated": "2021-04-30T16:44:57Z",
  "depositor": "",
  "mdtype": "",
  "split": {
    "requestDate": "2021-04-30T16:44:57Z"
  },
  "_attachments": {
    "metadata": {
      "content_type": "application/octet-stream",
      "revpos": 5,
      "digest": "md5-YQl+v66c2vfiuI0yKK7nkA==",
      "length": 66204,
      "stub": true
    }
  }
}
```


I also decided to always use "Content-Type": "application/octet-stream" , as it doesn't matter to the back-end code what mime type is used for the attachment.   This avoids some interesting potential complexities with sending the file, as the server needs to know what type of data might be PUT to it.


https://github.com/crkn-rcdr/sapindale/blob/5de02e3b39e636867e1ec5725d6ab14c57971201/src/server.js#L20-L23


Now that everything is in the same repository, part of this work will also be to port the concept from Sapindale.


https://github.com/crkn-rcdr/sapindale/blob/main/src/routes/dmd.svelte
https://github.com/crkn-rcdr/sapindale/blob/main/src/components/DMD/Editor.svelte

SaschaAdler:
See #172.

PR #65, opened 2021-02-16
view changes for collection

RussellMcOrmond:
Some changes to views, and possibly schemas due to the upgrade to CouchDB 3.


RussellMcOrmond:
There is a _view/slug in both `collection` and `manifest` databases.

One possibility that has been discussed is to create a new database schema based on merging these two types, and adding a third for slug aliases.  This way a Mango index can be used on a common slug field that will find any document (of any of these types) with a given slug.

Once a schema has been defined, tools can be written which migrate documents from the existing databases to the new database.

PR #68, opened 2021-03-06
@crkn-rcdr/access-data: Data validation package

SaschaAdler:
The first thing that struck me as odd about working on accessor was that I'd have to redo data validation between it and the schemas in `couchdb/schemas`. This is my attempt at fixing it.

Some changes will need to be made to Kivik for this to work with `kivik validate`, but the good news is I'll actually be stripping functionality from Kivik for this to work. See https://github.com/crkn-rcdr/kivik/issues/47 for that work.


SaschaAdler:
@RussellMcOrmond Apologies, but we need to invoke kivik slightly differently now (hopefully with only one extra letter). Instructions are in the README.

If you get it to work, I'll merge this PR.

PR #73, opened 2021-03-15
Creating new metadatabus related views/indexes for `access` database

RussellMcOrmond:
Related to https://github.com/crkn-rcdr/CIHM-Meta/issues/35

* Hammer2 doesn't need much, and I believe the existing views and update document will just work.
* For `makepublic` I'll need to be able to look up slugs, which I assume will be an index and `/_find` rather than the old `_view/slug`
* For `loadseries` it currently loads all existing collections to determine what is an update and what is new.  An index on the `type` field and using `/_find` will offer that.


RussellMcOrmond:
It there are comments on these, let me know.  I expect that the Date.parse() issue is resolved, but don't know if  this is the best time to change the view as it would require that we change the related Sapindale tools which use those views.  All the metadata bus cares about is that the updateinternalmetaq view is sorted in date order.

RussellMcOrmond:
I believe the thinking with slug is that we still want to have a view, so we can pass in an array of slugs to look up in one query.

I note with /_explain that there is an `mrargs`, which makes me wonder if it is possible to somehow pass in mapreduce arguments.

The concept with _find seems to be that there is a mapreduce and then the selector is used to filter the remaining results.  It feels like this is a start, and that being able to pass in regular map/reduce parameters plus point /_find at a /_view would complete this functionality.


In the interim, how do we want to complete the views/etc we need at the moment, recognising that a future version of CouchDB might make some views redundant?

I want to tackle Smelter and `loadseries` next.

* Smelter currently uses `/manifest/_design/access/_view/slug` to look up a single key= at a time (to avoid slug collisions).
* `loadseries` uses `/manifest/_design/access/_view/slug` to load all slugs and create a hash, as well as loading all the collection documents.

I'm curious what you think about an index on the type field for when we want to separate these, or if we should be using a /_view where we can do multiple lookups, have quick access to counts, and use include_docs when that is appropriate.

SaschaAdler:
@RussellMcOrmond I'm trying to figure out where/how the `noupdateinternalmeta` and `updateinternalmetas` views are used. I understand what they're for, overall, but not why they emit what they emit.

RussellMcOrmond:
> @RussellMcOrmond I'm trying to figure out where/how the `noupdateinternalmeta` and `updateinternalmetas` views are used. I understand what they're for, overall, but not why they emit what they emit.

They can be cleaned up.

I think this was entirely visual to have specific values output when doing curl/etc on the command-line..  The /_find interface replaces all of that, and will be especially powerful if your suggested enhancement is implemented.  That and I tend to write things in perl now rather than bash -- which is a bad reason to have written things they way they were..

SaschaAdler:
I'll remove them. I don't think I see code that's currently using them.

RussellMcOrmond:
> I'll remove them. I don't think I see code that's currently using them.

We'll need the views in the future for tools, and will be used via CouchView in the interim.

https://admin.canadiana.ca/couchview/manifest/metadatabus/updateinternalmetas?group=2&keytype=json&reduce=true

(This is also where the value is currently used by staff -- I kinda forgot).


At the moment there is nothing else that will indicate to staff when there is a metadatabus failure, and these views are what we'll be using (What ID's have never been run, which had failures, etc).

SaschaAdler:
Yeah, we'll need something that displays current errors. I'll update how that works.

SaschaAdler:
There's a lot going on in my latest update.

1. A whole bunch of things that seem useful throughout were extracted into functions. Once I figure out how to get Kivik to handle CommonJS imports, I'll add them to a "prelude" library that every function can read from.
2. The `extractData` function that will be part of the prelude no longer looks at `req.form`. I can change that back if necessary, but I think `Role::REST::Client` can handle JSON requests without too much trouble. The only thing that uses `extractData` right now is `_design/metadatabus/_update/hammerResult`.
3. The contents of the `basic` update are now spread across four different places, corresponding to the four different operations that `basic` provides, which presumably never get run all at once. The logic for how they operate hasn't changed too much, I don't think.
4. Views have been renamed and their logic cleaned up. `_design/metadatabus/_view/failures` (should be `hammerFailures`, will fix that shortly) emits our favourite date array only when Hammer fails. `_design/metadatabus/_view/neverProcessed` replaces `noupdateinternalmeta`.
5. I've left `_update/basic` and `_view/updateinternalmetaq` in place, although their functionality is superseded and they can be deleted when nothing is using them.

One thing I've done in cleaning up various bits of logic is assume that everything in the database matches our spec. In the case of these design doc functions, I'm not checking if `doc.updateInternalmeta.requestDate` exists if `doc.updateInternalmeta` exists. I want anything where that condition isn't met to loudly bring everything to a halt, because it means something invalid has entered our database.

I've tested this locally, but I haven't deployed it yet. Let me know if I should.

SaschaAdler:
Made the changes we discussed in our chat, and deployed this to iris.

PR #82, opened 2021-04-28
Admin tools

SaschaAdler:
Separate from sapindale for ease of integration with other code changes. We'll see how this goes.

Won't be ready to go until Docker workflows are in place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 **/node_modules
 **/dist
 **/build
-pnpm-debug*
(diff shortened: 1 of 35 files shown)


SaschaAdler:
Working:

- a local dev environment (no auth required)
- Docker, although the new sveltekit tools require a development build in the Docker image, for now

Tomorrow: Accessor, and then I think we can merge this.

SaschaAdler:
Accessor is working.

PR #134, opened 2021-05-18
Port slug tools from sapindale

DivyaKrishnan26:


SaschaAdler:
I'll have a longer look at this tomorrow, but as a quick note, if we add the `components` directory to the `lib` directory, we can take advantage of some path-rewriting later on, such that in our routes,

```ts
import { Component } from "../../../components/foo";
```

becomes

```ts
import { Component } from "$lib/components/foo";
```

services/admin/src/lib/components/TypeAhead.svelte
```diff
@@ -1,35 +1,34 @@
...
   import { createEventDispatcher } from "svelte";
 
   const dispatch = createEventDispatcher();
 
   export let label = "Please provide a label for this component.";
 
-  let prefix = "";
-  let lookupList = [];
+  let query = "";
+  let lookupList: string[];
```
DivyaKrishnan26, line 9: I have added the variable to be in TS but i still get an error or that, and I couldn't find any solution to it. 
That is why i have left other variable assignment to the JS notation.

DivyaKrishnan26:
I feel there is a bit of caching issue or the event is getting triggered weird, The datalist responds a bit unstable. 
This is from my testing.

services/admin/src/lib/components/TypeAhead.svelte
SaschaAdler: I'm not sure what error you were getting, but TypeScript is happy with the current iteration of this.

SaschaAdler (approved):

SaschaAdler:
I've made a couple of changes to things:

1. I sorted out a few issues with the TypeAhead component with TypeScript, and fixed the new URL in one of the fetch calls.
2. You had been using an older version of the slug resolver code. I've updated that to the latest version and sorted out TypeScript as well.

I'm ready to merge this, but let me know if you'd like to make any changes first.

DivyaKrishnan26:
We can go ahead and merge this(I had the URL changed when making changes already yesterday).
The next thing would be is to work on the layout, and bringing in the editor with values.

Issue #83, opened 2021-04-29
Staff update tracking


SaschaAdler:
Currently, when an access object is edited by staff (which doesn't actually happen yet, but bear with me), `updateInternalmeta.requestDate` is updated to cause the metadata bus queue to process the object. It would be nice to keep track of which staff member performed this edit, so that any problems with the metadata bus processing can be directed to that person specifically. In the future we will also want to create these queues for multiple processes per object, or for back-end processes on other data (e.g. `dmdtask`).

At first I thought that adding staff information to a generic `updated` field was the best approach for this, but now I'm wondering if we just want to add staff information along with `requestDate` in the [`ProcessUpdate`](https://github.com/crkn-rcdr/Access-Platform/blob/main/data/src/util/ProcessUpdate.ts) specification. Although, again, per #72, we need to keep track of staff updates generically, and not all updates will request changes in these queues. We could always track both:

```json
{
  "updated": {
    "date": 1234567.89,
    "by": {
      "name": "Sascha Adler",
      "email": "sadler@crkn.ca"
    }
  },
  "updateInternalmeta": {
    "requestDate": 1234567.89
  }
}
```

but I wonder if we'd rather attach the staff member to the generic update or the queue request (or both??).

@RussellMcOrmond curious for your input here.

RussellMcOrmond:
`updateInternalmeta` is itself ephemeral in that it will be used only as long as we have an `internalmeta`.

It almost feels like there should be a separate field for when we know a `by`, as I don't see value in storing the "name" of a service that makes an update.

At this point my recommendation would be a third field that looks likw what you suggested above (date and by), but that is separate from `updated`.

I recommend we don't think of this as tracking, but a reminder.  I don't think keeping an audit trail or anything similar would be worth the work.

SaschaAdler:
`updateInternalmeta` has a shelf life but I'm hoping the ProcessUpdate specification that it uses can last longer, which is why I bring it up.

Do processes like Hammer and whatever other daemons we put together need to update `updated`, if they also update the ProcessUpdate they're tied to, and especially if they don't change anything in the object itself? Do we want an `updated` field that only keeps track of human intervention?

We could certainly specify that `by` is optional, or we could come up with some scheme to say "Smelter created this", or we could even keep track of who imported something in Smelter and save that to the new objects Smelter has created.

RussellMcOrmond:
I've added @brittnylapierre and @DivyaKrishnan26 to this issue.

[...]

PR #139, opened 2021-05-19
dipstaging / staff update

brittnylapierre:
Hey @SaschaAdler I can make any changes you need me to make based on what's there now. I'm not sure if there's other issues or stuff to work on that has to do with this change but I can do that too. Thanks!


RussellMcOrmond:
Can I request that the title for this be more descriptive?   Thanks :-)

SaschaAdler:
The only comment I have on this is that I'm not sure a METS object needs its own dedicated type; I don't know if it's going to show up anywhere else. This certainly looks like it does what it's supposed to, and since I'm trying to think up solutions to #143 before I sign off for a while I think we can just merge this as is.

SaschaAdler (approved):

SaschaAdler:
Related to #83, although I wouldn't mind using this in practice before closing that issue.

PR #141, opened 2021-05-19
Admin tools: get an object for editing

SaschaAdler:
As advertised, except it doesn't work yet.


SaschaAdler:
There are a couple of problems:

1. `data` depends on ajv, which means that client-side code that uses `data` type-guards depends on ajv. It almost certainly shouldn't.
2. ajv and vite aren't getting along. It's possible that svelte-kit and/or vite updates help with this.

SaschaAdler:
The kit/vite update solved the ajv problem, somehow. Separating the types from the schema stuff is still important, but it's way less urgent, and we can definitely do it in the data package.

Merging this now. Can't wait for this to become more useful than it currently is. :)

PR #147, opened 2021-06-03
Manifest editor

brittnylapierre:
Hey @SaschaAdler! Let me know if I should add comments to the code I have written to improve readability for your review. I wasn't sure if I should for the front-end. Hopefully the structure I chose for the components makes sense too. I can change anything/make suggested improvements. Thanks!


brittnylapierre:
Actually I feel like this should be a draft still because the back end work will  need to get done eh before we close the issue? @SaschaAdler

services/admin/src/components/canvasmanifests/EditorActions.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+  import type { CanvasManifest } from "@crkn-rcdr/access-data/src/access/CanvasManifest";
+  // fastest object operators?
+  import equal from "fast-deep-equal";
+  import { cloneDeep } from "lodash";
```
SaschaAdler, line 5: This looks like an import that will cause the entire lodash library to be copied into the site bundle. I can think of two alternatives:

1. see if there's an import like `import * as cloneDeep from "lodash/cloneDeep"`
2. use something like [rfdc](https://www.npmjs.com/package/rfdc) instead

services/admin/src/routes/edit/[noid].svelte
```diff
@@ -0,0 +1,126 @@
+<script lang="ts">
+  import "../../app.css"; // TODO: Is this the best way to implement global styling?
```
Comment (SaschaAdler) [inline services/admin/src/routes/edit/[noid].svelte:2]: How about linking to it directly in the HTML template? We could store it in /static.

SaschaAdler (changes requested):
I'm impressed with the polish and attention to detail this interface already demonstrates. I have a few major requests, and some comments attached to particular bits of code as well.

1. I think it might be possible to restrict the viewport from allowing vertical scrolling for the editor as a whole. If individual tabs, or components inside individual tabs, need to stretch out longer than the viewport, they can do so themselves, like you how have the canvas thumbnail list working. Right now, to see the bottom of the last canvas in the thumbnail list, I have to scroll both it and the page itself.
2. Your point in slack about using Svelte actions for some of the DOM manipulation first made me look up what Svelte actions are, exactly. Once I did that, I agree that they might be helpful. I haven't followed the flow of your code enough as of yet to have specific recommendations for where, though.
3. I think the editor will end up being referenced from the /object/[prefix]/[noid] route, which might be a useful place to store the "Side Menu" layout, and then decide which of the tabs to display based on what kind of object we're looking at.

brittnylapierre:
Thank you Sascha! These reviews are so helpful, I'll get started on the changes to make today 🙂

SaschaAdler:
Two issues I'm having with the drag-and-drop:

1. If I drag something and let it go outside of the list, the list ends up in a transitory state. (I'm not sure what's changed in the underlying data, if anything.)
2. When dragging an item on top of another item, the list doesn't update the new order until after the item is dropped. This means that it looks like you're duplicating the item on top of its new location, instead of shifting the other items to accommodate it. The outcome at the end is correct, but I wonder if this will be confusing.

I'm having a hard time attaching the videos I took with examples of the two issues to this comment; I'll send them to you on Slack.

SaschaAdler:
I've looked around a bit more and I have a couple of comments, but overall I think we're ready to merge things.

1. I'm not sure if using components to abstract away CSS is the right approach, especially when the components have implicit defaults. I'm thinking specifically of the Align component, which I don't think gives much more value than just providing the relevant styles to its parent element. I might want to do a more thorough survey of how global styles and such are set up; I think we want to use something akin to [CUBE CSS](https://cube.fyi/), where Svelte components implicitly provide the scoping in the Block layer.
2. We should move the actions and components directory under `lib`, because that way our imports can look a lot cleaner. (I'd do this ASAP.)
3. I'm not sure if the SideMenu/Toolbar stuff should be in every routes' layout, but it certainly should be refactored into a layout that routes can inherit from. There should be a way to provide a property to routes that sets the toolbar title and list of panes. This will help coordinate porting over collection editor stuff from Sapindale.

I'm really looking forward to showing this to people outside the team!

brittnylapierre:
@SaschaAdler Oh cool, I'll look at CUBE CSS thank you! 

I pushed a commit just now with potential fixes for what you saw with the drag and drop as well. 

I'll let you know once I've fixed the 3 comments above and we can merge? 

I can wait and have you do item #1 since I'm less familiar with best practices in that area if you'd like. Or, I can read up on it and copy the best practices from their site, whatever you prefer!

Thanks again for your feedback!

SaschaAdler:
Don't worry about the CSS stuff for this PR. I think we'll need to apply whatever system we come up with across the whole site.

brittnylapierre:
Hey @SaschaAdler! For tomorrow, I pushed possible fixes for the drag and drop issues, moved the components under lib, and tried to clean up the editor route page.

Let me know if you have any suggestions especially for the editor route page cleanup. I'm not super happy with it yet, but I think I might need to do more thinking or reading to make it work exactly like how you described. I'll check out the Sapindale repo again to see how it's done there tomorrow morning too!

I left some comments in the editor page file with my question-y thoughts.

New possible route for the editor page would be: http://localhost:3060/object/69429/edit/manifest/m02n4zg6h671

Then I think http://localhost:3060/object/69429/m02n4zg6h671 would just be used to display the object.

brittnylapierre:
Hey @SaschaAdler I think this is the best version I can think of for now for the editor page. Let me know if you have any suggestions. I have some thoughts/questions as comments in the route files.

Link for editing a manifest: http://localhost:3060/object/69429/edit/m02n4zg6h671
Link for a collection: http://localhost:3060/object/69429/edit/s0vq2s46j98h

SaschaAdler:
I think I'm going to merge this shortly, and then do layout work in a new PR. One thing I'll do beforehand is move the drag-and-drop demo off of the front page.

SaschaAdler (approved):
Any further work on this can go in their own issues.

PR #152, opened 2021-06-09
Design doc prelude

### LINKED ISSUE #85: Extract generic design doc functionality to a prelude
Kivik now supports CommonJS modules in `lib` views. The functions that I've copy-pasted throughout design docs can be extracted to a single file, and symlinked into any design doc directory that needs it.

SaschaAdler:
I'm confident this works. Resolves #85.

@RussellMcOrmond Let me know if the way this works is clear. We can talk through it sometime if you like.


RussellMcOrmond (approved):
Once I read https://docs.couchdb.org/en/3.1.1/query-server/javascript.html#commonjs-modules , then this made sense.

I suggest that this link go somewhere a new developer would see as part of on-boarding/etc, or adequate comments added to the prelude so it is obvious what this relates to.

PR #157, opened 2021-06-18
Add canvas to manifest menu layout and CSS cleanup

brittnylapierre:
Adding a draft pull request for the work on cleaning up the CSS and adding the add canvas menu


brittnylapierre:
Hey @SaschaAdler I made some changes based on the CSS reading you sent & hooked up the add canvas menu to Divya's component. I'll have to ask you and her what the best way will be to filter the results table on 'isManifest' or any other type when we get a chance to speak again.

Let me know if you want anything changed thanks!

SaschaAdler:
A couple quick things to start:

1. In the `main` branch I moved the editor route into `/object/[prefix]/[noid]`. I don't think we need a separate `edit` URL. (By the way, the only reason `prefix` and `noid` are set up that way is because our IDs have a slash in them. They're part of the same thing, conceptually.)
2. Something in the CSS math has changed and there's a bit of unused layout at the bottom of the page. Not a big deal but I'm not sure where it's all set up.

I probably have a lot of thoughts about the CSS changes, but I'm not sure if I'll get the time to dive deeply into them.

brittnylapierre:
ah okay - I figured you could probably base the CSS layout you'll make on this branch instead, the main change is using rem instead of PX.

Oh right I have one improvement to make for that - I wanted to allow for the side menu to be 'inline' or 'full page' based on a property value. What you're seeing is if it's inline, so still part of the page flow. 

I'll push a change for it to have the option to be either, make the editor page full page, then delete the edit route

brittnylapierre:
Okay I made the page full screen again and deleted the /edit route. I can make any other changes needed, thanks!

SaschaAdler:
I have some minor quibbles but I think it's worth moving past them for now. I'll create an issue for one thing in particular.

I'm ready to merge this, if you are.

SaschaAdler (approved):

PR #161, opened 2021-06-25
Improve editor layout

brittnylapierre:
Hey @SaschaAdler I'm not sure if there's an issue to go along with this but feel free to attach it!


services/admin/src/lib/components/shared/SideMenuContainer.svelte
```diff
@@ -50,28 +75,62 @@
...
-<div
-  bind:this={container}
-  class="side-menu-container"
-  class:fixed-full-page={fullPage}
->
-  <slot name="side-menu-header" />
-  <div class="auto-align auto-align__a-stretch ">
-    <slot />
+<div bind:this={container} class="side-menu-container">
+  <div class:fixed-full-page={fullPage}>
```
SaschaAdler, line 79: Something here is confusing GitHub's syntax highlighting.

services/admin/src/lib/components/shared/SideMenuContainer.svelte
```diff
@@ -1,10 +1,35 @@
...
+  export let pageList: Array<SideMenuPageData> | undefined = undefined;
+
+  let pageNames: string[] = []; // Makes it easier to generate the sidenav buttons list
+  let pageComponents: any[] = []; // Makes it easier to generate the page body list
+  let instances: any = {}; // Keeps track of the actual instances of the components dynamically created throughsvelte:component.
+
+  $: pageNames = pageList ? pageList.map((el) => el["name"]) : [];
+  $: pageComponents = pageList ? pageList.map((el) => el["componentData"]) : [];
+  $: {
+    for (let key in instances) {
```
SaschaAdler, line 24: Purely stylistic thing, but if a variable is never going to change, I'd recommend using `const`. It's rare in my experience that `for (let key in keys)` is preferable to `for (const key in keys)`.

services/admin/src/lib/components/shared/SideMenuPageListButton.svelte
```diff
@@ -4,13 +4,12 @@
 
 <style>
   li {
-    background: var(--grey);
-    filter: brightness(1.1);
+    background: rgba(0, 0, 0, 0.06);
     padding: var(--perfect-fourth-7) var(--perfect-fourth-4);
```
SaschaAdler, line 8: I just noticed the musical ratio thing and I like it.

services/admin/src/routes/object/[prefix]/[noid].svelte
```diff
@@ -1,65 +1,87 @@
...
+  import EditorActions from "$lib/components/access-objects/EditorActions.svelte";
+  import StatusIndicator from "$lib/components/access-objects/StatusIndicator.svelte";
+  import InfoEditor from "$lib/components/access-objects/InfoEditor.svelte";
 
   let object: AccessObject;
   const objectStore: SvelteStore<AccessObject> = getContext("objectStore");
   object = $objectStore;
   setDataModel(object);
 
+  let pageList: Array<SideMenuPageData> | undefined;
```
Comment (SaschaAdler) [inline services/admin/src/routes/object/[prefix]/[noid].svelte:18]: I'd instantiate `pageList` with `const pageList: Array<SideMenuPageData> = [];` and then push items into it. It'll help make the logic below less repetitive.

We might want to create a type guard for whatever I called the parent type of Collections and Manifests. (I wish I used traits instead. Maybe someday.)

services/admin/src/routes/object/[prefix]/[noid].svelte
```diff
@@ -1,65 +1,87 @@
...
+
   let rfdc: any; // Deep copies an object
   let model: AccessObject;
```
Comment (SaschaAdler) [inline services/admin/src/routes/object/[prefix]/[noid].svelte:21]: This is the unchanged copy, right? I'm not sure `model` is the best name for this; it's unclear what it's a model of.

services/admin/src/routes/object/[prefix]/[noid].svelte
```diff
@@ -1,65 +1,87 @@
...
-      <span slot="editor-content">
-        <!--SideMenuPage overflowY="hidden">
-        Other Content! {JSON.stringify(object)}
-      </SideMenuPage-->
-      </span>
-    </EditorLayout>
-  {/if}
+  <div class="editor">
+    <SideMenuContainer {pageList}>
+      <Toolbar slot="side-menu-header" title={object["slug"]}>
```
Comment (SaschaAdler) [inline services/admin/src/routes/object/[prefix]/[noid].svelte:76]: Not something to solve with this PR, necessarily, but we might want to extricate the Toolbar from the SideMenuContainer, for use on pages where we want it but no SideMenu.

SaschaAdler (changes requested):

brittnylapierre:
Thanks for your feedback Sascha! I'll make the required edits.

brittnylapierre:
@SaschaAdler I pushed fixes from the review! 

I couldn't make the pageList const otherwise it wouldn't draw the menu but I made it have the default value of []. 

I changed the variable model to objectModel and added a comment explaining it is where the changes to the object go before the user presses save. 

I added a bit more templating for the app colours (made the bg of the canvas viewer match what's on canadiana.ca now, so the app currently has more of a light theme.) Feel free to play around and change the defaults! It's pretty easy to make something like a dark mode now. If people want to clone this repo it should be easy for them to re-skin it with their own colours too. 

For the InfoEditor.svelte, I'm not sure if we should pass an object or if it would instead be better to have multiple of props (ex: slug, label, ...) that can be null, and show props that are instantiated only? If it matters, if you have a preference I can apply it to all components.

Issue #143, opened 2021-05-19
Ensure schema validator is only instantiated when necessary


SaschaAdler:
I'm running into problems in #141 where vite is having a hard time loading ajv, when I know full well it shouldn't have to. It seems like we're really going to need a sandbox package containing TypeScript types and a few code utilities like type guards, that we can compile to ESM without worrying.

I'd say this is a pretty high priority as I don't think we've seen the last of these problems.

Effectively this means splitting the schema validation code from the TypeScript data model, and I'm not sure how we should do that quite yet.

SaschaAdler:
The kit/vite update in #145 solved the urgent ajv import problem. This is still an important thing to sort out, but it's way less urgent, and we have more options available since if we can figure out how to depend on ajv some but not all of the time, we might be able to keep everything in the package it's currently in.

SaschaAdler:
A more recent kit/vite update for #160 is causing problems with this again. Instead of creating a new package, I think it might make more sense to have validation take place in `accessor`, while still keeping the schema in `data`. I'll sort out a more detailed plan for this in the next few days.

SaschaAdler:
What I'd like to do with this:

1. Experiment with https://github.com/bcherny/json-schema-to-typescript to avoid having to rewrite our schemas into TypeScript (or vice versa, but going from TypeScript to JSON Schema is lossy).
2. `data` will only contain schemas, types, and type guards. It will also contain a utility method for adding all of the schemas to an Ajv instance, but Ajv won't be a dependency.
3. `accessor` will depend on Ajv directly. `couchdb` and `lapin` will depend on it for schema validation.
4. `accessor`, and not `data`, will handle conversion between CouchDB documents and the data that `accessor` and our services work with directly. I'm not sure if a) we need schemas for CouchDB documents, and if so, b) where they should live. I'd like to avoid having them in `data` since `data` is supposed to be universally useful.

SaschaAdler:
Potential thought: set up `couchdb-driver` package for translating between Couch and everything else. If this model works, we can also set up a `swift-driver` and whatever else we'd like. The `couchdb` service can rely on `couchdb-driver` for testing.

SaschaAdler:
Closed in the work last week.

PR #162, opened 2021-06-28
Dev environment update

### LINKED ISSUE #86: Production Docker image for the admin tools
Keeping track of workspace dependencies locally means that Docker images can't be built just by copying some files in a particular service's directory. A more comprehensive solution is required.

### LINKED ISSUE #160: Bump svelte-kit
Will allow us to specify route parameters to avoid having to assert their `string`-ness.

SaschaAdler:
Resolves #160, #143.


SaschaAdler:
Quick note on this: there's a bug in Vite (https://github.com/vitejs/vite/issues/3940), or SvelteKit's use of Vite, that causes vite to throw an exception when Svelte throws a CSS error. The CSS error is correctly reported in VSCode.

(In our case the error was that the Svelte `:global` CSS directive can only accept one selector at a time.)

PR #165, opened 2021-06-28
Collection Editor-First version

DivyaKrishnan26:
Contents In the viewer, Pending the Member list and reordering the member list


DivyaKrishnan26:
Few Questions from my side,(For the Member List)
1.For the memberlist, do we want to do something similar to the canvas list displays?
2.If yes, does all the manipulations(delete members, add members, reorder members) apply? Publish and unpublish will apply I guess.
3.For the display, does a grid suffice or do we want to have thumbnails, Although I am modifying to keep it as thumbnails so far.

DivyaKrishnan26:
I guess this PR and the branch can be deleted. I can do it if its alright with @SaschaAdler   and @brittnylapierre

PR #166, opened 2021-06-29
"Deleting" objects in their editor - branched off of improved editor layout

brittnylapierre:
Hey @SaschaAdler I thought we may be able to use the concept of "placing the object in storage" to mean removing it from whatever it's a member of and putting it away in a safe place, where it can be viewed and added back into the platform later if people want to.

We could also use: archive, trash, recycle bin, bin...

Let me know what you think, thanks! 

(Basing this off of the bigger editor layout fix branch, so I'll merge master back into this later once there's no issues with it and it has been merged to master)


RussellMcOrmond:
On the "placing object in storage", please see the discussion in https://github.com/crkn-rcdr/CIHM-Meta/issues/12 

There is a specific field in a canvas document to indicate it is orphaned (this can happen more ways than staff removing from manifest).  We have discussed creating specially named manifests which include orphans and other things.

RussellMcOrmond:
Removing a canvas from a manifest is removing an element from an array.  No need to worry about the thing you removed from the array any (the canvas record in its database)  further at this point.  We can get fancy in Accessor later (IE: trigger the "am I an orphan" functionality, etc).

At this point we only need to replace the functionality we are breaking with not allowing SIP updates and splitting identifiers (AIP IDs vs Slugs).


Same thing with removing a member from a collection -- it is at this point only an array manipulation, and we don't need to concern ourselves with the details of the member we have removed.

Issue #158, opened 2021-06-23
Drag-and-drop mid-drag status


SaschaAdler:
Not a priority, but I think the drag-and-drop system would work better if the underlying list is edited when an item is dragged into a new spot, and then edited back if the re-ordering is cancelled by moving the item back into its original position, or off the list entirely.

The empty space that dragging leaves in the original spot is difficult to process visually, I think, although it's better than leaving the original item in its place.

PR #171, opened 2021-07-13
Basic UI features & Code from moved from /admin to /proof-of-concept (with required updates for changes in main)

brittnylapierre:
Included in this branch are:

**1. Basic styling of the application**

The styles.css contains style code based off of tutorials found here: https://piccalil.li/tutorials/ 

**2. Drag and drop functionality improvements**

As per #158 the dragged item is moved through the array as the user drags it.

**3. 'Deleting' objects in their editor.** 

There is a storage bin button at the top right of the editor, that allows the user to press it to move the object into storage. Then, a modal comes up saying:

Are you sure you want to place this object in storage?
By placing this object in storage you will be taking it out of all the collections it belongs to. You will be unassigning its slug, 'oocihm.8_06941_1.' You can then use that slug for other objects. Objects that do not have a slug assigned to them are effectively undiscoverable. You can bookmark this page to access this object again in the future. You can assign it a new slug to make it discoverable in the platform again.

The user presses a 'place in storage' button to confirm and actually unassign the slug. 

**4. Creating new objects from a blank editor**

When the user navigates to: http://localhost:4747/object they see a screen saying 'Create a new:' with buttons for 'Manifest' and 'Canvas'. Then, when the user presses one of the buttons, a blank editor displays with the appropriate sections for each object type.

**5. Code from moved from /admin to /proof-of-concept + basic back end code added to lapin-router to get all of the old requests working again**

I am reading the couch docs then will read couch-util to help me make anything I did in lapin-router better and help me out when writing the save button (my next todo.)


package.json
```diff
@@ -22,6 +22,7 @@
     "cleanup": "pnpm run kivik:stop",
     "dev": "run-s down dev:up",
     "dev:up": "docker-compose --profile dev up --build --remove-orphans --detach",
+    "dev:up:watch": "docker-compose --profile dev up --build --remove-orphans",
```
SaschaAdler, line 25: One of the benefits to starting docker-compose in detach mode is that you don't have to worry about accidentally hitting Ctrl-C and bringing the whole thing down.

Maybe we should add a `watch` script which just runs `docker-compose logs -f`.

packages/lapin-router/src/routes/slug.ts
```diff
@@ -1,19 +1,56 @@
...
+        ? { noid: response?.[0]?._id, slug: response?.[0]?.["slug"] }
+        : null;
+    },
+  })
+  .query("resolveMany", {
+    input: Array,
+    async resolve({ input: q, ctx }): Promise<[Noid, Slug][]> {
+      const response = await ctx.couch.access.find({
+        selector: {
+          slug: { $in: q },
```
SaschaAdler, line 42: Sadly, using `$in` causes CouchDB to scan through the entire index, instead of picking out the values in the provided array, which is why we have a JavaScript view set up to do this. That said, we should just add something in couch-utils that simulates what this should be doing.

SaschaAdler (approved):
Just a couple of things worth pointing out. I think we should merge this ASAP regardless of the slug resolve comment that will need to be sorted out.

I'm not going to look into all of the admin tools changes but feel free to point out anything in particular you want me to look at.

brittnylapierre:
Thank you Sascha! I'll make issues for your two comments and paste your explanations in them  :)

PR #185, opened 2021-07-23
Separating new and edit existing object pages

brittnylapierre:
I moved the editor related code to a sub component and cleaned up the following routes:

Access-Platform/services/admin/src/routes/object/[prefix]/[noid].svelte
Access-Platform/services/admin/src/routes/object/new.svelte
Access-Platform/services/admin/src/routes/object/__layout.svelte

This setup makes it easier for us to show a loading status for the editor page and in general will reduce confusion if for whatever reason the object isn't fetched from the database correctly.

Having the common code between the pages in a component helps for re-use.

You should be able to search for objects from the front page, click their row, see a Loading... string then the editor should pop up.

You should be able to click new object from the front page, click either manifest or collection, and see a blank editor for the object type you selected.


SaschaAdler (approved):
Big fan of pulling this logic out of the layout component, if nothing else.

PR #186, opened 2021-07-23
couch-utils updates

### LINKED ISSUE #182: Some couch-utils functionality
Useful things include:

- [ ] `find` a single record where a field matches a string (or more?); return `null` if nothing is found; throw if more than one record is found
- [ ] using the above method, `find` records for a list of strings, [since we can't use `$in` like we could have expected](https://github.com/apache/couchdb/issues/1853)
- [ ] utility that returns `${str}\uFFF0` when provided `str`
- [ ] `find` everything where a field starts with a given string, using that utility
- [ ] generic update function caller
- [ ] throw away `updateWithSubset` in favour of update functions, which is what we should probably stick to

I might look into TypeScript generics that will allow `find` results to be typed based on whatever `fields` are requested.

SaschaAdler:
Closes #182.


SaschaAdler:
I'm not sure that the TypeScript generic stuff I'm playing with here is worth it, and comments are definitely welcome.

SaschaAdler:
Note to self:

Do the nano-like thing, but one better, and assign a Zod parser to each db handle. Maybe figure out how to do it such that things like `_id` and `_attachments` lose the underscore.

You can then use the associated type to control what fields are looked up by `find` requests.

PR #174, opened 2021-07-19
Back-end hookups for saving objects

brittnylapierre:
To be included in this pull request:

- [x] Improvements made to the validator of noid.resolveMany 
- [x] Save route added with validation
- [x]  Confirmation status animations added for back end save request (note for B - make some sort of service were backend calls can hook into optionally as the developer writes them to show result status information)
- [x] Validation checks as the user edits objects in the editor
- [x] Save button disabled if the object is in an invalid state

PR #196, opened 2021-07-27
Adding deleting members

DivyaKrishnan26:
This should have the Drag and drop of members adding member elements and deleting the member elements.

**To Clarify:** I have added the Noid as the new member to the list to Make the link got to the corresponding editor( as I remember we doing the same thing in sapindale). I would want to double confirm on that if we would still be wanting to that or something different in that case.

**Thing to take forward with**: Create the member link to view in the editor for either a Collection/ Manifest.

Note: the Stylings are all kept similar to the Canvas editor.


SaschaAdler:
I'm having a hard time reviewing this code because I'm running into the same problems with this branch that I ran into with #174. Unfortunately the code that solves this for me is in that branch.

DivyaKrishnan26:
Apologies for making it hard enough, let me know if I can merge the #174 branch into this branch if that might help you.

SaschaAdler:
Not your fault! You could try rebasing this to #174 if you like.

DivyaKrishnan26:
Sure I will try that.

SaschaAdler:
I've merged main (with a few edits). Since we're going in a different direction with lapin on how updates happen, the update code you wrote no longer works. I'll let you know when the new version is ready.

PR #198, opened 2021-07-27
Admin tools documentation and response handling

brittnylapierre:
I'm adding comments that will load on hover for our components and modules, giving an overview, listing properties, and showing example usage.

This way, it will be easy for the non-authors of components to know how to use them. 

At the same time I'm taking care of any formatting issues of imports that were automatically added that don't use $lib.

You should be able to see documentation about the svelte components when you hover over them in vscode. See: https://github.com/sveltejs/language-tools/blob/HEAD/docs/README.md#documenting-components

When merging main into this branch for testing, I found I needed to edit the components that called the backend in places due to the change in response format from lapin. That has been done as well in this PR.


SaschaAdler:
You've got some empty temp files committed in this branch, I think.

brittnylapierre:
@SaschaAdler Thanks! I've removed them, I'm not sure how they were generated

SaschaAdler:
Sounds good. I'll look into this more on Monday.

SaschaAdler:
This looks good.

I think some of the comment annotations might be redundant since we're using TypeScript in our components; I don't believe you need to specify types with `@type` when you're also specifying them in the code itself, for example. Not an important point.

Are we using all of the utility components still? Ones like `Align`.

brittnylapierre:
Ahh thanks @SaschaAdler I saw it somewhere else in the code so just went with that format. We arent using the Align component, I'll audit the rest to see what I can delete and then push again!

brittnylapierre:
Okay! @SaschaAdler @DivyaKrishnan26  I merged everything in and this branch is ready to be reviewed. I added some TODO documentation comments to new components just for the sake of not holding things up for @DivyaKrishnan26 to get started on the bulk add feature.

PR #203, opened 2021-07-30
database-specific handlers

### LINKED ISSUE #202: `couch-utils` database-specific handlers
Based off of discussion in #200, which says, minus the alternatives that I won't go with:

---

My intention when I set up `couch-utils` is that it would be a wrapper around `nano` that would hide away a lot of the stuff in `nano` that we aren't using, as well as make life a little easier when using `nano` with TypeScript. (`nano`'s typings are incredibly strict, not always complete, and get in the way more often than they help, in my experience.) The idea would be that `DatabaseHandler` doesn't have any functionality that only applies to the fields we expect in one particular database, and that it could be used with any Couch database. That said, the only place we can perform database-specific functionality is in `lapin-router`, right now, and that might make router `resolve` functions unwieldy.

I propose that `couch-utils` should contain a `handlers` directory where `DatabaseHandler` is extended for each database, when necessary. `couch-utils` will also provide the entire `couch` context object that lapin uses. Here are a couple of examples of how we might want to do this:

Simple example (no need for database-specific functionality): The `object.get` procedure's `resolve` method calls `ctx.couch.access.get(id)` to get the access object reference by that id. `get` is already defined in `DatabaseHandler`.

Complex example (need for database-specific functionality): For `object.unassignSlug`, the `resolve` method calls `ctx.couch.access.unassignSlug(id)`. The `access` handler has defined this `unassignSlug` method, in which it looks up if the object with this `id` is a member of any collections by using the `members` view, calls the `removeMember` update function on each collection the view lookup returns, and then calls the `unassignSlug` update function on the object with `id`.

This model doesn't help with any operations that need to use multiple database handlers, but I don't think those will come up too frequently and they can be handled in `lapin`.

SaschaAdler:
Closes #202.


SaschaAdler:
This is just about ready to go; I'm going to make sure that when a slug is being edited `editObject` checks that it isn't in use elsewhere.

brittnylapierre:
Hey @SaschaAdler this looks really clean and easy to understand!

Issue #113, opened 2021-05-05
Descriptive metadata bulk upload interface


SaschaAdler:
An interface that accepts descriptive metadata files, uploads them to newly created `dmdtask` documents, and causes the right background processes to run.

A front-end proof of concept exists in sapindale, but will likely need to be updated to conform to changes in #111. A lot of this should be integrated into accessor as well.

RussellMcOrmond:
A [document was written](https://crkn.sharepoint.com/:w:/s/platform/EW2AmkBTXZZEnwWWSAsIHDcBlRWSAYf9gL3oKPrvr4EU0A?e=vIIbN2) in August 2020 to describe the thinking at the time.  What we are doing is simpler than that, as we don't need a complex API to carry out this simple task.


* multi-record file format uploaded (format needs to contain an ID in record) from web interface
* back-end process splits multi-record format into separate validated XML records, storing details in Couch document
* Web interface used to confirm ID's, and pick subset to be loaded.
* back-end would copy the XML files to the right place.


A very basic demonstration of the upload form was already added to Sapindale, which can be ported or at least used as inspiration #55

RussellMcOrmond:
https://github.com/crkn-rcdr/sapindale/issues/97  provides additional description, but given that ticket is in the context of Sapindale it can now be closed.

RussellMcOrmond:
I uploaded the 3 .mrc files that are in [Sample Metadata](https://crkn.sharepoint.com/:f:/r/sites/platform/Shared%20Documents/Development%20-%20Current%20Projects/Access%20Platform1/Sample%20Metadata?csf=1&web=1&e=EctkHu) so far to the production database (View via Iris).

Using the `dmd-task-swift` branch, and running `pnpm prod` pointing to the production database..


https://access-dev.canadiana.ca/dmd/517ee635a2305bc78b58b9efc1264a07

It was success with 0 items, not what I expected. I forgot which one this was -- There is a file that I hadn't tested yet, so I'm guessing it was that one.

* A new feature I suspect would be useful is to store the filename of the file that was uploaded, so the operator will know which metadata file they are looking at the results for.  Some of these may be long running processes, and there may be more than one tab open waiting for results.


Next is https://access-dev.canadiana.ca/dmd/517ee635a2305bc78b58b9efc126723a which on the browser just says "Loading...", but there is an error message in a 500 error from the server-side of Access ("invalid_union", etc).
[...]

Issue #100, opened 2021-05-05
Import interface (Smelter)


SaschaAdler:
Staff will use an admin tool interface for selecting packages to import into the access platform as manifests. Staff can look up package records in `dipstaging` (through accessor logic that needs to be built) by providing a list of slugs (or also by date?). Packages with slugs that are not already in use can be imported.

Related `dipstaging` work is in #99.

SaschaAdler:
There is a functioning tool for this called "Archival Manifest" in sapindale. I think I now finally understand what the name "Archival Manifest" actually means, finally: it's from the archive, as opposed to created inside the access platform. Right?

I like "Import from Preservation" better but whatever works for staff works for me.

RussellMcOrmond:
My recommendation is to use the term "AIP ID" or similar and not "Slug" when referencing records in `dipstaging`.

Those IDs look sorta like slugs at the moment because of the current tie between Preservation and Access platforms, but in the future the AIP ID's will be very different.

See:  https://www.archivematica.org/en/docs/archivematica-1.12/user-manual/archival-storage/aip-structure/

> For example, looking at the AIP name my-aip-d31cc44f-ce01-4e67-affe-513868d9cf3d, my-aip is the name assigned by the user and d31cc44f-ce01-4e67-affe-513868d9cf3d is the UUID generated during ingest.

RussellMcOrmond:
> There is a functioning tool for this called "Archival Manifest" in sapindale. I think I now finally understand what the name "Archival Manifest" actually means, finally: it's from the archive, as opposed to created inside the access platform. Right?

It is a term that William came up with during a much earlier meeting.

Thoughts for other developers:

Those concerning themselves with user experience should think about what the best terminology will be for staff using these tools. Same with the term "Ingest" and "Import", to ensure the distinctions are clear.  It all makes sense to me, but what I think isn't generally the best indicator of what other staff will be thinking :-)   Sounds like good things to add to the Glossary.md

A reminder when porting Sapindale tools to Access-Platform:  Sapindale was written assuming that the web client would be directly talking to CouchDB using https://github.com/crkn-rcdr/upholstery .  CouchDB and other databases and object storage is now being abstracted by Accessor, so no need to think in the older mindset.

To make the connection, the back-end tool which processes the updated `dipstaging` records  is called **Smelter**.

SaschaAdler:
[...]

PR #206, opened 2021-08-03
Landing page format change and scaffolding for high priority features added

brittnylapierre:
I've added the links for the #100 and #113 features and moved the 'Create Object' into the navbar too. The user has a help icon with a link to our issues page, and can click the user icon for user related information (right now just the session text.)

We can add NotificationBars as needed to the front page, and the user can search for collections and manifests to edit too.

I added the scaffolding for the two big features discussed today, #100 and #113 that are navigated to via the links in the nav bar.


brittnylapierre:
@SaschaAdler I've fixed some more editor issues and saving existing manifests/collections, so if you want we can merge this now/let me know if you want anything changed.

There are two other things I will need to fix, that I think can done in new pull requests.

The first is saving new objects - I wanted to know if you wanted me to make a route for this and if so if you have an implementation in mind?

Second is setting slugs to undefined for 'putting objects into storage' - if I pass an empty string, the edit fails, if I pass null for the slug, it fails too. If I could pass in undefined I know it would work, but every time I set data to: { slug : undefined } the slug property gets trimmed out and I get an error about the data object being empty. I'm not sure if you already know a solution for this but if not we can work on a fix in a new branch, too!

Thanks! (fyi @DivyaKrishnan26 )

SaschaAdler:
A quick skim of the code looks fine; if you want to merge now, go ahead.

As for the two other things, they will both need new routes.

1. I was too hasty with setting `EditableManifest` and `EditableCollection` to be partial types, although I think the solution there is to leave them as partial and then use `NewManifest` and `NewCollection` for the non-partial versions. The new routes would then accept e.g. a `NewManifest` object, mint a noid for it, and then call an AccessHandler method that calls an update function that only succeeds if the noid isn't in use (it shouldn't be, but who knows). The update function will also need to know the type of the object; I can help with its implementation. A note there is that while `slug` is technically optional, the new object interfaces should require one (and it should be unique).
2. This will also require a special route and shouldn't be something you can do by zeroing out the `slug` input in the editor. There should be an AccessHandler method that accepts a noid (and the User?) and
    1. Looks up the noid in the `members` view
    2. Runs an update function on each Collection found in the lookup that searches through the Collection's members list and removes the entry with the noid as the `id`, then updates `updateInternalmeta`
    3. Runs an update function that removes the slug from the object identified by the noid and updates `updateInternalmeta`

As an aside we need to add "updates `updateInternalmeta`" to the design doc prelude, since basically every update function does it.

brittnylapierre:
Nice, okay! I'll merge this then I 'll get started on the docs branch. I'll create issues for items mentioned above and work on them as I finish them too. Thanks Sascha!

PR #209, opened 2021-08-05
Porting existing sapindale DMD task components into admin to use as a base for the DMD form UI. 

brittnylapierre:
I've taken the template code from: https://github.com/crkn-rcdr/sapindale/tree/main/src/components/DMD 
And added the new components to our project under: /services/admin/src/lib/components/dmd with some edits since we've changed the architecture of the app.

Three utility components were created to facilitate form functionality (use of which may change now that I have a better understanding of the requirements:)
-  /services/admin/src/lib/components/shared/FileSelector.svelte
-  /services/admin/src/lib/components/shared/ScrollStepper.svelte
-  /services/admin/src/lib/components/shared/ScrollStepperStep.svelte

Slight changes were made to the navigation around the app.

Based on discussions today, I will edit the design of, and work on the functionality for, the /dmd/new page and the /dmd/:id page in later PRs


SaschaAdler:
You've got a bunch of spurious files committed into this branch. They have "Pull request #209" in their filenames?

brittnylapierre:
Ugh I'll have to double check for this every time I open prs, I'm not sure how they're being created! Thanks @SaschaAdler

brittnylapierre:
Actually, i don't see them on my machine even when I pull my branch again I'll look if they're only visible in github?

brittnylapierre:
Okay! The files that were generated somehow have been deleted

SaschaAdler:
I suspect the schema updates are going to change a lot about how this interface works. Do we need `ScrollStepper`? There are only two steps in which users will be able to interact with things, and they don't really follow the "stepper" paradigm, especially since one step will be at a "new" route and one at a "task" route.

We could merge this now, regardless, but I think it might be worth holding on to reevaluate the interface based on #230.

brittnylapierre:
@SaschaAdler Exactly I want to just merge this to start off with, but I am going to change the design based off of my better understanding of what needs ti happen, including not having the ScrollStepper. 

I just wanted to get this in as a starting point, then I'll start working on each individual step once your dmdtask branch is merged too

SaschaAdler:
Sure thing. Happy to merge now, then.

PR #213, opened 2021-08-06
Couch utils testing

### LINKED ISSUE #184: couch-utils testing
Update: I'd rather get this working first for couch-utils. I'll add another dev service for watching it (and other working tests), but some things may not be ported for a while.

SaschaAdler:
Closes #184.


SaschaAdler:
This is ready to go.

Note a **BREAKING CHANGE** in the repo's filesystem, which is that kivikrc.json has moved. Steps you'll need to take:

1. `path/to/repo:$ cp services/couchdb/kivikrc.json ./kivikrc.json`
2. Add the `subdirectory` field to `kivikrc.json` as done in `kivikrc.example.json`

This lets us run kivik anywhere in the repo and it will use the config in `services/couchdb`.

RussellMcOrmond (approved):
I can only comment on the comment -- was that intended to be a "mv" or a "cp" of the kivikrc.json that is currently in services/couchdb/ ?

I suspect the intention is to only have one, and for it to be in a parent directory that would allow it to run from any child directory.

SaschaAdler:
Whoops, yeah, that's supposed to be a move.

PR #215, opened 2021-08-09
Unassigning slugs back-end

### LINKED ISSUE #159: Filter slug lookups based on type
The canvas-adding flow in #157 uses the generic slug lookup to find manifests to add canvases from. We don't currently have a way to filter the lookup to only return manifest slugs.

### LINKED ISSUE #211: Create back-end code for removing slugs ('putting objects into storage')
This will require a special route and shouldn't be something you can do by zeroing out the slug input in the editor. 

There should be an AccessHandler method that accepts a noid (and the User?) and:

- [ ] Looks up the noid in the members view
- [ ] Runs an update function on each Collection found in the lookup that searches through the Collection's members list and removes the entry with the noid as the id, then updates updateInternalmeta
- [ ] Runs an update function that removes the slug from the object identified by the noid and updates updateInternalmeta

### LINKED ISSUE #212: Add "updates updateInternalmeta" to the design doc prelude
We need to add "updates updateInternalmeta" to the design doc prelude, since basically every update function does it.

SaschaAdler:
- [x] Closes #159 
- [x] Closes #211 
- [x] Closes #212 
- [x] part one of #216

### DIFF
--- a/packages/couch-utils/src/handlers/access.spec.ts
+++ b/packages/couch-utils/src/handlers/access.spec.ts
@@ -88,6 +88,23 @@ test.serial("Cannot edit something if it's the wrong type", async (t) => {
   t.true(error.message.includes("has type: manifest"));
 });
 
+test.serial("Can unassign a slug", async (t) => {
+  await t.context.access.unassignSlug({ id: MANIFEST_TWO, user: USER });
+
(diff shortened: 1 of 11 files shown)


DivyaKrishnan26 (approved):

SaschaAdler:
Check it out, I'm doing actual test-driven development.

Issue #227, opened 2021-08-10
Dmdtask: Copy XML to Access and/or Preservation platform storage.


RussellMcOrmond:
See [Documentation in Sharepoint](https://crkn.sharepoint.com/:w:/s/platform/EW2AmkBTXZZEnwWWSAsIHDcBlRWSAYf9gL3oKPrvr4EU0A?e=Xq5kvk) for details.

The last stage of the process is to copy the set of XML files into the platforms.

* For the existing Preservation platform, these are dmd.xml attachments to the documents in the `wipmeta` database.
* For the new Access platform, these are dmdTYPE.xml files stored in Swift

As there is no further processing of the XML at this point, only storage, we have the option of doing this work within `admin` and `lapin` rather than doing it in Perl.

Note to @SaschaAdler  :  You moved the storage policy file via https://github.com/crkn-rcdr/Access-Platform/commit/0e5ca5d8654cb907566e33e3fc9c053246e1b8a9 , and then removed Accessor.  We probably want to get that documentation back  :-) .  I wanted to link to it in this issue, which is why I noticed it went missing.

SaschaAdler:
See #231.

brittnylapierre:
Hey @RussellMcOrmond Sascha told me to implement the label portion for the access update. Right now, the label gets updated in the "access" database when someone updates the metadata in access. Is there a label key for the object in wipmeta that needs this to be updated too? Or is it just handled by the update to dmd.xml? Thanks!

RussellMcOrmond:
I made my notes in a different ticket https://github.com/crkn-rcdr/Access-Platform/issues/259#issuecomment-915209832

Yes, label is part of both access and preservation.

PR #233, opened 2021-08-10
DMDTask schema updates

### LINKED ISSUE #232: `dmdtask` schema updates
See discussion in #230.

---

- Remove references to `store` processes as lapin will take care of that
- Remove the `prefix` field, as the interface will be configured to look up with a prefix after the split.
- Make `id` optional in `SplitRecord`, in case the reason the result isn't valid is that an id couldn't be extracted.

SaschaAdler:
Closes #232.

@brittnylapierre I'm not sure if your code references the schema directly yet; I've changed a few names for things.

I'll be creating some examples of what these should look like in `dmdtask` fixtures.


brittnylapierre:
@SaschaAdler Yeah! It uses a dmdtask object that is manipulated in the DmdEditor but it's okay if it breaks what I have now, I'll need to edit it based on our plan. If my pr is merged, then this one is merged, I'll create a tiny pr for fixing what i have now to work with the schema changes, then subsequent prs for each step outlined in: https://github.com/crkn-rcdr/Access-Platform/discussions/230 I think! that way I keep things focused and small

SaschaAdler:
I've added some test data that should parse. When we start working on couch/lapin we can test it out as fixtures.

brittnylapierre (approved):
I like the new schema changes it's easy to understand the states that the split task can be in. Thanks Sascha!

RussellMcOrmond (commented):
"    "0.json": jsonAttachment,"

I believe the intention was for this json to be part of items[]?

I like the tests, as it gives me a visual example to confirm I'm understanding what the zod is saying..

RussellMcOrmond:
While adding a note to #227 227 I realised I forgot about labels.  Each metadata format has its own place for labels, which is why it is part of the output.

https://github.com/crkn-rcdr/CIHM-METS-App/blob/64ef1c43697e37815f1fee344bfb4fc5d22c61c4/lib/CIHM/METS/App/Marc.pm#L105-L110


We will want to expand our  ParseRecord to include the label, and to decide what to do with that label.  Current tool just imposes it, so we can start with that and see if staff ever want it optional or editable later.

RussellMcOrmond:
* Up to you for the JSON.  We can keep as an attachment.  I just thought you were thinking of moving it into the ParseRecord in the above.  It becomes a Lapin question if the full record, or only a filtered subset, is sent back to other layers.  This is mostly an optimisation question, ans since these CouchDB documents are ephemeral we can change later if we want without a data migration.

* Label isn't included in the JSON -- the JSON is the parsing of the XML, and the Label is separate.  In the case of the CSV format the label is a separate column (like the ID, which is also not in DC or IssueInfo).

* We should go with the default now, and consider these things as enhancements.  Once staff is actually using the new platform there will be many things that will be more clear than our currently extremely abstract conversations.

PR #234, opened 2021-08-11
Batch member addition

DivyaKrishnan26:
Notes:
I have marked my doubts in the files as comments. You might not want to review the svelte files yet. Its just the scaffolding stuff to get the structure away.


packages/couch-utils/src/handlers/access.ts
```diff
@@ -105,6 +118,18 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
+
+      const filteredMembers = xorWith(data.members, currentMembers, isEqual);
+
+      console.log("data in accesshandler", filteredMembers);
+      await this.forceUpdate({
+        id: args.id,
+        user: args.user,
+      });
+    }
+    /*Return the Collection Object here? after forceUpdate?*/
```
SaschaAdler, line 132: The Collection is retrieved again and returned at the end of this method.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -105,6 +118,18 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
+    if (data.members) {
+      const currentMembers = Collection.parse(await this.get(args.id)).members;
+
+      const filteredMembers = xorWith(data.members, currentMembers, isEqual);
+
+      console.log("data in accesshandler", filteredMembers);
+      await this.forceUpdate({
+        id: args.id,
+        user: args.user,
+      });
```
SaschaAdler, line 130: This needs to be fleshed out a bit. It's the members that you've filtered that need to have `forceUpdate` run on them.

SaschaAdler (changes requested):
If you'd like, I could help you write a test case that covers what the collection member check needs to do.

packages/couch-utils/package.json
```diff
@@ -70,6 +72,9 @@
     "@crkn-rcdr/access-data": "workspace:^0.0.0",
     "@crkn-rcdr/access-env": "workspace:*",
     "http-errors": "^1.8.0",
+    "lodash": "^4.17.21",
+    "lodash-es": "^4.17.21",
+    "lodash.xorwith": "^4.5.0",
```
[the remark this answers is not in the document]
DivyaKrishnan26: Deleted the other lodash and types

packages/couch-utils/src/handlers/access.ts
```diff
@@ -63,7 +65,18 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
-
+  /**
+   * forceUpdate an Access Object.
+   */
+  async forceUpdate(args: { id: Noid; user: User }) {
+    await this.forceUpdate({
+      id: args.id,
+      user: args.user,
+    });
+  }
```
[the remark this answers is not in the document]
DivyaKrishnan26: So we would want to call the update function here, makes clear

DivyaKrishnan26:
Yeah, It would be good to know what the collection member checks need to be done exactly.

packages/couch-utils/src/handlers/access.ts
DivyaKrishnan26: Which means this goes in like 
await this.update ({ 
id:args.id,
filteredmembers may be?
})

packages/couch-utils/src/handlers/access.ts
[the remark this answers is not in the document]
DivyaKrishnan26: Doesn't the update method calls the DatabaseHandler method?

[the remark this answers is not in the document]
SaschaAdler: There's no DatabaseHandler `forceUpdate` method.

You can call the DatabaseHandler `update` method to call the `forceUpdate` update function in Couch. It should look like

```ts
this.update({
  ddoc: "access",
  name: "forceUpdate",
  docId: id
})
```

This `forceUpdate` method doesn't need a `User` passed in, either, so it can just have the signature

```ts
async forceUpdate(id: Noid): Promise<void> {
  ...
}
```

packages/couch-utils/src/handlers/access.ts
SaschaAdler: For each member in `filteredMembers`, you'll need to call `this.forceUpdate`.

SaschaAdler: Oh, shoot. You also need to do this after `editObject` gets called below, since otherwise Hammer might not pick up on the changes.

SaschaAdler:
I've merged main into the branch (a couple times, apparently), and added a test case that currently fails, but will succeed when members are updated.

Use `pnpm watch packages` to watch the test output. You might need to wait a minute or two for all the tests to run first.

packages/couch-utils/src/handlers/access.ts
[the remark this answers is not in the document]
DivyaKrishnan26: Should there be any return type for the Promise? Like if the members should be returned here?

packages/couch-utils/src/handlers/access.ts
DivyaKrishnan26: The forceUpdate method take in just the Id and the Promise right? So call the forceUpdate(id) for each of the filteredMembers ?

DivyaKrishnan26: Just trying to figure out how the forceUpdate should be called or each filteredMembers based on the function decleration.

SaschaAdler: forceUpdate just takes an id (the Promise bit in the type declaration is there because it's an `async` function). You'll need to call it for each member's `id`.

packages/couch-utils/src/handlers/access.ts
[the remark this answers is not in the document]
SaschaAdler: A function that returns nothing has a return type (i.e. what goes after the colon) of `void`. Since this is an asynchronous function, it has a return type of `Promise<void>`. If you remove the `Promise` you should see an error.

`forceUpdate` should be an asynchronous function that returns nothing.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -113,30 +113,38 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
     data: EditableCollection;
   }): Promise<Collection> {
     const data = EditableCollection.parse(args.data);
+    let membersArray: string[] = [];
     if (data.members) {
       const currentMembers = Collection.parse(await this.get(args.id)).members;
 
       const filteredMembers = xorWith(data.members, currentMembers, isEqual);
 
       console.log("data in accesshandler", filteredMembers);
```
SaschaAdler, line 129: You'll want to remove this before merging.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -113,30 +113,38 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
-      for (const members of filteredMembers) {
-        await this.forceUpdate(args.id);
+      //filteredMembers.map((members) => members.id);
+      for (let members of filteredMembers) {
+        if (members.id != undefined) {
+          membersArray.push(members.id);
+          await this.forceUpdate(members.id);
+        }
       }
```
SaschaAdler: You have an array of members in `filteredMembers`. You're pushing their ids into a different array, and then running `forceUpdate` on each id in that array. Why use two arrays? There's a type in `access-data` called `ObjectList` that can help here.

```ts
let filteredMembers: ObjectList = [];

if (data.members) {
  filteredMembers = xorWith(data.members, currentMembers, isEqual);
}

// await this.editObject

for (const member of filteredMembers) {
  await this.forceUpdate(member.id);
}

// return, etc.
```

The way you're setting it up, you're running `forceUpdate` on everything twice, which, while it isn't a problem, is unnecessary.

SaschaAdler (changes requested):

packages/couch-utils/src/handlers/access.ts
DivyaKrishnan26: I had the other array as Typescript was complaining about assigning values that are of different types. Did not thoroughly know about the ObjectList[], I will make the changes now.

SaschaAdler (approved):

PR #235, opened 2021-08-11
Unasssigning slugs, publishing objects, and searching for only manifests in the add canvas menu.

brittnylapierre:
Fixes the front end /admin app in response to: https://github.com/crkn-rcdr/Access-Platform/pull/215
- Slugs can be unassigned (object placed in storage)
- Objects can be published/unpublished
- Add canvas menu only brings up manifests in the search.

### DIFF
--- a/services/admin/src/lib/components/access-objects/EditorActions.svelte
+++ b/services/admin/src/lib/components/access-objects/EditorActions.svelte
@@ -81,8 +81,8 @@ The editor actions component holds functionality that is responsible for perform
           return e;
         }
       },
-      `${objectModel.type} created!`,
-      `Failed to create ${objectModel.type}.`
+      `Success! Created ${objectModel.type}.`,
+      `Error: couldn't to create ${objectModel.type}.`
     );
   }
 
@@ -117,8 +117,8 @@ The editor actions component holds functionality that is responsible for perform
           return false;
         }
       },
-      "Changes saved!",
-      "Failed to save changes."
+      "Success! Changes saved.",
+      "Error: failed to save changes."
     );
   }
 
@@ -181,8 +181,8 @@ The editor actions component holds functionality that is responsible for perform
         }
         return false;
       },
-      `Placed ${objectModel["type"]} in storage. (Slug unassigned!)`,
-      `Couldn't place ${objectModel["type"]} in storage. (Slug not unassigned!)`
+      `Success! Placed ${objectModel["type"]} in storage. (Slug unassigned!)`,
+      `Error: couldn't place ${objectModel["type"]} in storage. (Slug not unassigned!)`
     );
   }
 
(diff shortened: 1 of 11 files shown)


services/admin/src/lib/components/access-objects/EditorActions.svelte
```diff
@@ -243,27 +274,26 @@ The editor actions component holds functionality that is responsible for perform
...
-    slug to make it discoverable in the platform again.
+    By unassigning this object's slug, you will be taking it out of all the
+    collections it belongs to. You can then use the slug, '{object["slug"]}' for
+    other objects. Objects that do not have a slug assigned to them are
+    effectively undiscoverable. You can bookmark this page to access this object
+    again in the future. You can assign it a new slug to make it discoverable in
+    the platform again.
     <!--You'll be able to view
     {object["slug"]} in storage and add it back into the platform
     <a href="/storage" target="_blank">here.</a-->
```
SaschaAdler, line 289: This comment is out of date, right?

services/admin/src/lib/components/access-objects/EditorActions.svelte
```diff
@@ -200,7 +191,47 @@ The editor actions component holds functionality that is responsible for perform
...
+              objectModel["public"] = undefined;
+            } else {
+              const response = await $session.lapin.mutation(
+                `accessObject.publish`,
+                {
+                  id: objectModel.id,
+                  user: $session.user,
+                }
+              );
+              objectModel["public"] = Date.now() / 1000;
```
SaschaAdler, line 219: I'm confused by the difference between `object` and `objectModel`. Which one represents what's on the server?

Assuming it's `objectModel`, it's safer to re-retrieve the object from the server any time anything changes, instead of editing fields that are also edited on the server. For example, the `public` field below is probably different from the one the server sets.

services/admin/src/routes/index.svelte
```diff
@@ -13,7 +13,7 @@
    */
   function slugSelected(event: CustomEvent<string>) {
     const noid = event.detail;
-    goto(`/object/${noid}`);
+    if (noid) goto(`/object/${noid}`);
```
SaschaAdler, line 16: Try using `Noid.parse` or `Noid.safeParse` here.

SaschaAdler (changes requested):
Just a few small things I noticed.

Feel free to create a new one for re-retrieving the object from the server after changes to it if it's hard to implement in this one.

services/admin/src/lib/components/access-objects/EditorActions.svelte
brittnylapierre: @SaschaAdler object represents the server state - objectmodel represents the editor state. Maybe I can rename object to serverObject and objectModel to editorObject?

I'll edit the code to grab the object from the server again after any change (including slug unassignment and publishing.) I can make a method in editor actions for 'pullObjectFromServer()'

We have both objects to keep track of changes, so that we only show the save button if anything has changed from the user input. (This lets us hide the save button it if they change an edited field back to what it was before.)

Let me know if those two improvements (renaming the variables and pulling from the server again) are a good solution for your concerns or if you'd like me to do anything else!

Thanks

services/admin/src/routes/index.svelte
brittnylapierre: Ah thanks!

services/admin/src/lib/components/access-objects/EditorActions.svelte
SaschaAdler: Those improvements work for me.

services/admin/src/lib/components/access-objects/EditorActions.svelte
```diff
@@ -166,30 +164,23 @@ The editor actions component holds functionality that is responsible for perform
...
-            //error = e;
             console.log(e);
           }
         }
         return false;
       },
-      "success",
-      "fail"
+      `Success! Unassigned the slug '${objectModel["slug"]}.'`,
+      `Error: unassign the slug, '${objectModel["slug"]}.'`
```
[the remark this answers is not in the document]
brittnylapierre: I had to resolve this in a way that's a little less straight forward, but I pass the error message to the notification bar ad the 'details' On the notification bar, they see: Error unassigning slug '${editorObject["slug"]}.' & the user can click the info sign on the failure notification to see the details of the error message.

brittnylapierre:
@SaschaAdler I think I got all of the points above! I did end up implementing the pulling from the server I didn't see you said we could do it later :)

services/admin/src/lib/components/access-objects/EditorActions.svelte
```diff
@@ -81,8 +81,8 @@ The editor actions component holds functionality that is responsible for perform
           return e;
         }
       },
-      `${objectModel.type} created!`,
-      `Failed to create ${objectModel.type}.`
+      `Success! Created ${objectModel.type}.`,
+      `Error: couldn't to create ${objectModel.type}.`
```
SaschaAdler: "couldn't to create"

SaschaAdler (changes requested):
I think this might be it.

services/admin/src/lib/components/access-objects/EditorActions.svelte
brittnylapierre: Dang I can't read today! Thanks @SaschaAdler I fixed it

SaschaAdler (approved):

PR #236, opened 2021-08-11
Replace isequal from fast-deep-equals with lodash's single module for .isequal for canvas comparison

brittnylapierre:
I noticed with the other method that when I added the same canvases from: https://access-dev.canadiana.ca/object/69429/m02n4zg6h671 to itself (selected the first couple canvases) that the isequal from fast deep equals returned true, which is wrong. So I switched to using:

https://www.npmjs.com/package/lodash.isequal

I saw somewhere on github that we will make our own implementation, I'll switch to that once it's done.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -116,8 +116,8 @@ importers:
       '@types/jsonwebtoken': ^8.5.4
       cookie: ^0.4.1
       deep-object-diff: ^1.1.0
-      fast-deep-equal: ^3.1.3
       jsonwebtoken: ^8.5.1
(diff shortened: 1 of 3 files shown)


SaschaAdler:
We're not planning on using our own implementation for this, and so lodash is fine. I think the best approach for this on the front-end would be to use the `lodash-es` module, and then import `isEqual` with:

```
import isEqual from "lodash-es/isEqual";
```

since the `lodash.$FN` npm modules [are deprecated](https://github.com/lodash/lodash/issues/3838).

SaschaAdler (approved):

PR #237, opened 2021-08-11
Sidenav increase contrast between active and inactive tabs

brittnylapierre:
Maybe font colour can be used to indicate the active page? Any time I set the background to something too contrasting to the rest of the page my eyes go directly to the side nav instead of the editor fields.

On hover we can do some more fancy stuff, like lightening the background a fair bit.

### DIFF
--- a/services/admin/src/lib/components/access-objects/Editor.svelte
+++ b/services/admin/src/lib/components/access-objects/Editor.svelte
@@ -67,7 +67,7 @@ The editor component allows for the editing of AccessObjects. It will dynamicall
           },
         },
         {
-          name: "Content",
+          name: "Manage Content",
           componentData: {
             contentComponent: ManifestContentEditor,
             contentComponentProps: { manifest: objectModel },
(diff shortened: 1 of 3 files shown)


SaschaAdler:
That works too, I suppose. Mind showing a screenshot?

One question, as an aside: do we have consistent colour values for this kind of thing across components? I was expecting to see this background colour value as a CSS variable.

brittnylapierre:
@SaschaAdler  I need to figure out a tool for cropped screenshots on ubuntu lol I'll attach some soon (making this draft since changes will happen)

brittnylapierre:
On page load:
Screenshot from 2021-08-11 14-33-04

On hover of the active item:
Screenshot from 2021-08-11 14-36-18


On hover of a non-active item:
Screenshot from 2021-08-11 14-33-48

On manifest page:
Screenshot from 2021-08-11 14-34-16

SaschaAdler:
I think I like the last one.

brittnylapierre:
Okay! I could play around with the front end for literally ever so to avoid that I made an enhancement issue: https://github.com/crkn-rcdr/Access-Platform/issues/239 

That way we can work on it at some point but I don't feel the urge to spend too much time on it right at this moment :) 

I'll open this to make this the fix for it being really not noticeable between the two for now though

SaschaAdler (approved):

PR #241, opened 2021-08-11
couch-utils attachment methods, create dmdtask

SaschaAdler:
I might do some of the new DMDTask back-end stuff to test this out.


SaschaAdler:
I'm having a slow day, but here's this stuff taken care of, I think.

brittnylapierre (approved):
Looks good to me @SaschaAdler! Thanks

PR #243, opened 2021-08-12
Dmd Task: Step 1-3 Front-end

brittnylapierre:
### Interface: `/dmd/new`

- [x] File upload, which is base64-encoded to a string with [btoa()](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa)
- [x] Type of descriptive metadata (see below)
- [x] The admin tools should `goto` the route below.

Screenshot from 2021-08-12 13-48-48

### Interface: `/dmd/task/$uuid`

- [x] Display waiting state
- [x] Maybe tell the user how long it's been since they made the request?

Screenshot from 2021-08-12 13-48-30

## 3. Split operation failure

- [x] Show the error in `split.message`. 
- [x] Invite the user to upload a new file by going back to step 1.

Screenshot from 2021-08-12 13-52-05

### Documentation

- [x] Document the components

### DIFF
--- a/packages/lapin-router/src/routes/dmdTask.ts
+++ b/packages/lapin-router/src/routes/dmdTask.ts
@@ -1,3 +1,0 @@
-import { z } from "zod";
-import { User } from "@crkn-rcdr/access-data";
-import { createRouter, httpErrorToTRPC } from "../router.js";
--- a/services/admin/src/lib/components/dmd/DmdNewTaskForm.svelte
+++ b/services/admin/src/lib/components/dmd/DmdNewTaskForm.svelte
@@ -80,7 +80,7 @@ none
 <br />
 
 <div class="new-task-wrapper">
-  <h6>Start a new Metadata upload</h6>
+  <h6>Start a new Metadata Upload</h6>
   <NotificationBar message={errorText} status="fail" />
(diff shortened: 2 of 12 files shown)


SaschaAdler:
Whoa, that red is intense! I don't think that's a bad thing, necessarily.

I wonder if there's a preferred term for this stuff on the front-end than "DMD Task". Metadata upload? We don't have to change it yet.

brittnylapierre:
> Whoa, that red is intense! I don't think that's a bad thing, necessarily.
> 
> I wonder if there's a preferred term for this stuff on the front-end than "DMD Task". Metadata upload? We don't have to change it yet.

Lolol it's a hue shift of the Canadiana gradient ahaha! I wondered if people had the page up in the back of their computers, or on a second monitor, if changing the bg would catch their eye quicker! Let me know if you want a different gradient to be used, or a solid colour, or just white haha. I can quickly change it from saying DMD Task to Metadata upload! Honestly I'm just getting familiar enough with the process to understand what to say in each step now :')

brittnylapierre:
@SaschaAdler @RussellMcOrmond  pnpm run build successfully ran! I'll merge this now then you shouldn't have any trouble, sorry everyone!

PR #247, opened 2021-08-13
Schema updates and new ddoc functions for DMD processing

### LINKED ISSUE #246: Dmdtask:  Update CouchDB design documents (views and updates) to match current spec
Now that I believe https://github.com/crkn-rcdr/Access-Platform/blob/main/packages/data/src/dmd/Task.spec.ts is more firm, we can update https://github.com/crkn-rcdr/Access-Platform/tree/main/services/couchdb/dmdtask/design/access to match.

Needed by #113 and #112

SaschaAdler:
Closes #246.

I'm documenting in painstaking detail the things I'm doing as part of this PR.

---

1. I run `pnpm dev`, and then in two separate terminal windows: `pnpm watch packages` for building and testing, and `pnpm watch kivik` for design doc deploys to my local couchdb.
2. First, the schema changes in `packages/data`.
   - I make changes to `src/dmdtask/Task.spec.ts` with what I'd like for the new schema. I change `mdType` to `format` in `goodWaiting`. I take this opportunity to clean up the invalid tests at the bottom of the file, setting up a helper function to generate invalid objects. I also ensure that `output` is provided in each valid member of an `items` array. Saving this file breaks things.
   - `src/dmdtask/types.ts` contains the array with all possible values for what used to be known as `mdType` and will now be known as `format`.
   - I create a const array, `DMDFORMATS`, which contain each kind of format that metadata processing accepts. I create a TypeScript type, `DMDFormat`, using `typeof` to generate a string union out of the `DMDFORMATS` array. We can use these to ensure that nobody can assert a bogus value for the `format` field.
   - I do the same with `DMDOUTPUTS`.
   - At this point I also delete `MDTYPES` and `MDType`. This breaks more things.
   - I edit the schema in `src/dmdtask/Task.ts` to incorporate the new fields. This also requires editing the `refine` specification for `ParseRecord` because `output` has the same rules as `id` and `label`
   - Getting all this right caused tests to pass.
   - In order for the changes in `types.ts` to be exported correctly, I need to change both `src/dmd/index.ts` and `src/index.ts`.
3. Building packages shows that `couch-utils` has an error. It was expecting the `MDType` import to work. I quickly correct it to use `DMDFormat`. I also need to edit the `create` update function in `services/couchdb/dmdtask/design/access/updates/create.js` since the `mdType` field is now called `format`. I checked my `kivik` terminal to ensure that the design doc deployed. I then re-save the updated `dmdtask.spec.ts` to re-run its tests, and they pass. This is a good time to push a commit.
4. I now need to update the "queue" and "status" views for what is now only one metadata process. I can use the methods in the prelude to avoid copy-paste errors. Ironically, I'll just copy `hammerQueue` and `hammerStatus` from `access/design/metadatabus`, since they use the prelude methods already. I'll set the new views up at `processQueue` and `processStatus`. (This makes me wonder if I can pass the `emit` function as a parameter to a prelude method to do this sort of thing. Not to worry about now.) The new `ProcessUpdate` field is called `process`, and so I make sure to reference that. (As an aside, when I delete the other view files, kivik gives me an error, because I haven't implemented deleting files in watch mode yet. I think this can be safely ignored.)
5. You know, I should test that new DMDTasks show up in `processQueue`. Fortunately, I already have a test that creates a new DMDTask in couch-utils. I use the dmdtask handler's `view` method to get a response, making sure to pass `reduce: false` into the options because the view is set to reduce by default. I use `t.log(response)` to remind myself what view responses look like, and then write a test that matches the output. It passes. I don't have a proper way to test `processStatus` yet without setting up fixtures. Now's a good time to commit again. I also re-run `pnpm dev` to clean up all the extra views in `dmdtask` that are now gone.
6. I create a new update function called `process`, that the metadata processor will call when it's done parsing a metadata file. It seems to me that updating `ProcessUpdate` objects with their result is going to be something a lot of update functions do, and so I write a `processUpdate` method in `prelude.js`. I use this and other boilerplate functions (`successReturn`, `errorReturn`, `extractJSONFromBody`) to set the update function up. Since the data is coming in from a Perl tool that doesn't have direct access to the DMDTask schema, I also include some basic input validation inside the function. I check to make sure this function deploys correctly, and then delete `update.js`.
7. I had to fix a couple of things. See below.


RussellMcOrmond:
BTW: `processStatus` is more of a debugging view at this point.  I don't know if the web interface will want to output a list of status or not, but I doubt it.  This isn't like Hammer/etc where knowing failures is an important part of the UI.

RussellMcOrmond:
Some notes from my end -- not sure if helpful to put here...

(stop, re-run dev which starts fresh, watch logs, etc)


A reminder of how to set up the database if someone wants to test cihm-metadatabus things.   We set up production databases as the moment to not require the admin password except when creating/deleting databases/etc.  This means you need to allow users access without passwords.


```
russell@russell-XPS-13-7390:~/git/cihm-metadatabus$ ./russell-run 
...
tdr@3822b111979b:~$ dmdtask 
/dmdtask/_design/access/_view/processQueue?reduce=false&descending=true&limit=9 GET return code: 401
tdr@3822b111979b:~$ 
```


Go to https://couch-dev.canadiana.ca/_utils/#/database/dmdtask/permissions
Under Members/Roles there is "_admin" which requires anyone logging in to be a member of the _admin role.  Just hit the "x" beside that to remove that, and...

```
tdr@3822b111979b:~$ dmdtask 
tdr@3822b111979b:~$ 
```


Back in Couch I created a "new" minimal document for it to show up in the Queue...


```
{
  "_id": "fc6ebe10ac74ec255adbd43dff0002a8",
  "_rev": "1-51446ed90211c23cf7880e59179660da",
  "process": {
    "requestDate": "2020-02-02T02:02:02Z"
  }
}
```

And now my test works as I'd expect.

```
tdr@3822b111979b:~$ dmdtask 
$VAR1 = {
          '_rev' => '1-51446ed90211c23cf7880e59179660da',
          '_id' => 'fc6ebe10ac74ec255adbd43dff0002a8',
          'process' => {
                         'requestDate' => '2020-02-02T02:02:02Z'
                       }
        };
tdr@3822b111979b:~$ 
```

I don't know if any of the other developers will ever try to run the perl stuff on their desktops trying to talk to the local CouchDB, but if so I can provide details...

SaschaAdler:
@RussellMcOrmond I'm happy to test the `process` update function with dummy data in `couch-utils`, if you'd like.

RussellMcOrmond:
I might be misunderstanding how the update document is written.  It looks like there should be an array that has the 3 values.

But, I'm getting succeeded=false

```
  "process": {
    "requestDate": "2020-02-02T02:02:02Z",
    "processDate": 1628881452.996,
    "succeeded": false
  },
```

When I send in:

```
ProcessUpdate:
$taskid = 'fc6ebe10ac74ec255adbd43dff0002a8';
$updatedoc = [
               bless( do{\(my $o = 1)}, 'JSON::PP::Boolean' ),
               'Nothing to store here
Get to Marc soon
',
               []
             ];

$response = {
              'message' => 'Processed metadata for task fc6ebe10ac74ec255adbd43dff0002a8'
            };
tdr@841d7915d955:~$ 

```

SaschaAdler:
It's an object, not an array:

```ts
{
  succeeded: boolean;
  message: string;
  items: ParsedRecord[] | undefined;
}
```

I probably could have written it to not take `succeeded` at all, and then determine whether it succeeded based on the presence of `items`.

RussellMcOrmond:
Thanks.  Now I'm wondering where in the code that it stores items[] ?

I've only tested with ` 'items' => []` , but I looked and don't see where it is stored.

SaschaAdler:
I, er, forgot to set it. Give me a sec.

RussellMcOrmond:
Also: The way error returns are being done gets us into a form of an infinite loop.  If I get a $res->{error} I need to do another process update with false to ensure the same document doesn't just get processed over and over again.  Is that intended?

We need to decide what we mean by the errors, and how we want to respond to them.  All I can think of to do is to do another update with status=false and copy the message, but that is something you can do in the update (which I think is more clean by appending to the text of the message).

SaschaAdler:
Right, I should be successfully creating a failed task, then? That makes sense.

RussellMcOrmond:
FYI: My silly test is:
`        items     => [ { parsed => JSON::true, } ]`

I won't be finishing much of anything this evening for the tool, but want you to be able to merge this today as it will be out of mind by Tuesday/Wednesday :-)

SaschaAdler:
Although, I don't know. All of these errors should only happen if the processor isn't doing something it should be doing, instead of there being something wrong with the metadata itself.

If you could use (and heck, I could set it up with lapin) the schema to parse the update function's input, which is how we're ensuring that doesn't happen with JavaScript code, then I wouldn't even have put that validation stuff in.

RussellMcOrmond:
I agree that these are coding errors, but it is what we do when we have one of these.  Do we get stuck, or do we log an error and move forward?  The process should either be registered in the database as successful or failed, and this created a third state of the task not being removed from the queue.

Personally I'd rather the checks weren't being done in the update document at all.  These tests, if they are needed, should happen elsewhere.

SaschaAdler:
> FYI: My silly test is:
> ` items => [ { parsed => JSON::true, } ]`
> 
> I won't be finishing much of anything this evening for the tool, but want you to be able to merge this today as it will be out of mind by Tuesday/Wednesday :-)

If I were writing this in JavaScript I'd run a parser on the update input in couch-utils, which would fail your silly test :)

I think I want to merge this as is, although I could write some tests on the function with dummy data for you to work off of.

RussellMcOrmond (approved):
Core functionality tested..

PR #248, opened 2021-08-13
DMD task: Step 1-3 Lapin routes

brittnylapierre:
### Lapin route: `dmdTask.new`

Input:

```ts
type NewInput = {
  user: User;
  mdType: "csvissueinfo" | "csvdc" | "marc490" | "marcoocihm" | "marcooe";
  file: string;
};
```

Creates a new `dmdtask` document, through a couch-utils method, with an auto-generated uuid, the dmdType, and the file inserted as an attachment. Theoretically, since it's already base64-encoded, the `_attachments` object can be built in the initial request. Will have to research this a little bit. Decoding the string and then uploading it the standard way can work as an alternative.

- [x] Done

### Lapin route: `dmdTask.get`

Grabs the dmd task by id from the database

- [x] Done

### DIFF
--- a/services/admin/src/lib/components/dmd/DmdTaskTimeInfoTable.svelte
+++ b/services/admin/src/lib/components/dmd/DmdTaskTimeInfoTable.svelte
@@ -22,17 +22,17 @@ Displays a dmd task's time based information in a table, the user sees the task 
 </script>
 
 <div class="dmd-request-info">
-  {#if dmdTask?.["updated"]}
+  {#if dmdTask?.updated}
     <span>Request inititated:</span>
     <span>
-      {`${new Date(parseInt(`${dmdTask["updated"]}`) * 1000).toLocaleString()}`}
+      {`${new Date(parseInt(`${dmdTask.updated}`) * 1000).toLocaleString()}`}
     </span>
   {/if}
-  {#if dmdTask?.["process"]?.["requestDate"]}
+  {#if dmdTask?.process?.["requestDate"]}
     <span>Request updated:</span>
     <span>
       {`${new Date(
-        parseInt(`${dmdTask["process"]["requestDate"]}`) * 1000
+        parseInt(`${dmdTask.process["requestDate"]}`) * 1000
       ).toLocaleString()}`}
     </span>
   {/if}
--- a/services/admin/src/lib/components/dmd/DmdTaskTimeInfoTable.svelte
+++ b/services/admin/src/lib/components/dmd/DmdTaskTimeInfoTable.svelte
@@ -6,7 +6,7 @@ Displays a dmd task's time based information in a table, the user sees the task 
 ### Properties
 |    |    |    |
 | -- | -- | -- |
-| dmdTask: WaitingDMDTask or FailedDMDTask or SucceededDMDTask  | required | The dmd task to be displayed. |
+| dmdTask: DMDTask | required | The dmd task to be displayed. |
 
 ### Usage
 ```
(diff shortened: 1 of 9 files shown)


SaschaAdler:
We should probably set up type guards for DMDTasks, shouldn't we.

services/admin/src/lib/components/dmd/DmdTaskTimeInfoTable.svelte
```diff
@@ -6,34 +6,36 @@ Displays a dmd task's time based information in a table, the user sees the task
...
+    </span>
+  {/if}
+  {#if dmdTask?.["process"]?.["requestDate"]}
+    <span>Request updated:</span>
+    <span>
+      {`${new Date(
+        parseInt(`${dmdTask["process"]["requestDate"]}`) * 1000
+      ).toLocaleString()}`}
+    </span>
+  {/if}
```
SaschaAdler, line 38: Do you need to use `?.` and array-style indexing here? All of these fields are required in the DMDTask schema.

services/admin/src/routes/dmd/[taskId]/index.svelte
```diff
@@ -3,64 +3,86 @@
...
+    <DmdSplitSuccessStoreForm /> <!--dmdTask={getAsSucceededTask()} -->
   {:else}
-    <!--JUUUST In Case-->
+    <!--JUST In Case All Else Fails-->
     <DmdSplitFailureViewer
-      {dmdTask}
+      dmdTask={getAsFailedTask()}
       message="No objects were split from the metadata file."
     />
   {/if}
```
Comment (SaschaAdler) [inline services/admin/src/routes/dmd/[taskId]/index.svelte:88]: Same idea here has my other comment. I also wonder how Svelte handles the type guarding you get from using e.g. `WaitingDMDTask.safeParse`?

In regular old TypeScript, I'd do this with

```ts
if (WaitingDMDTask.safeParse(dmdTask).success) {
  // dmdTask is a WaitingDMDTask
} else if (FailedDMDTask.safeParse(dmdTask).success) {
  // dmdTask is a FailedDMDTask
} else {
  // because DMDTask is a union, you can be sure dmdTask is a SucceededDMDTask, even if TypeScript can't
}
```

SaschaAdler (changes requested):
I'm more curious about my comments than requesting that anything is changed, necessarily.

(even though I'm about to click "request changes")

services/admin/src/routes/dmd/[taskId]/index.svelte
Comment (brittnylapierre) [reply services/admin/src/routes/dmd/[taskId]/index.svelte:88]: Ooo that's much nicer, it looks like it works in the template too! I'll see if it gets rid of the type warnings so I don't need the getters too, thanks for suggesting this

services/admin/src/lib/components/dmd/DmdTaskTimeInfoTable.svelte
brittnylapierre: @SaschaAdler No i don't think so then! I remember in the last version of the page there were lots of checks like that but if the safeparse from below is enough checking then I'll just use that instead :)

brittnylapierre: @SaschaAdler  When I use dot notation, I do get some complaints for some of the properties, like: 
dmdTask.process.message

but I can do: 
dmdTask.process["message"] 

Without any issues, so I'll do that, and I can at least get rid of the ?. too now!

services/admin/src/routes/dmd/[taskId]/index.svelte
Comment (brittnylapierre) [reply services/admin/src/routes/dmd/[taskId]/index.svelte:88]: @SaschaAdler Looks like I still need to cast it if I want to be specific about the types allowed into the state-specific components :(

brittnylapierre:
> We should probably set up type guards for DMDTasks, shouldn't we.

@SaschaAdler I made some changes based on your above recommendations! I wasn't able to do everything perfectly though.

for the type guards, is that the same thing as isManifest() and isCollection() but for dd task types? I can add that too in this branch if you'd like!

brittnylapierre:
@SaschaAdler I added the isWaiting... typegaurds in my last commit, based on the ones you made and the safeparse suggestion from above! Let me know if there's any other things you'd like me to change, otherwise I'll merge this and rebase my next PR from main (up next, still working on it: https://github.com/crkn-rcdr/Access-Platform/pull/244) :smile:

SaschaAdler (approved):

PR #251, opened 2021-08-16
Swift

### LINKED ISSUE #245: Swift
- [ ] Set up a [picoswiftstack](https://hub.docker.com/r/swiftstack/picoswiftstack) container for dev and perhaps also local production
- [ ] Initialize the `swift-client` package
- [ ] Port some/all of [CIHM::Swift](https://github.com/crkn-rcdr/CIHM-Swift) into `swift-client`. We only need [object_put](https://docs.openstack.org/api-ref/object-store/?expanded=create-or-replace-object-detail#create-or-replace-object) and the auth stuff for the immediate use case
- [ ] Incorporate a SwiftClient into `lapin-router`

SaschaAdler:
Closes #245.

BREAKING CHANGE: The following are required environment variables:

- `SWIFT_URL`
- `SWIFT_USER`
- `SWIFT_PASSWORD`


RussellMcOrmond:
I'm assuming that "container" will become configuration variables for Lapin later?

I see the hint in the tests:

```
    accessFiles: client.container("access-files"),
    accessMetadata: client.container("access-metadata"),
```

SaschaAdler:
Yeah, the idea in lapin is that you'll be able to interact with objects in a particular container, but not do any of the scarier container management stuff.

The code above isn't in tests; it's in the `context` that lapin routers have access to.

PR #252, opened 2021-08-17
Dmd Task: Step 4 Interface

brittnylapierre:
### Interface: `/dmd/task/$uuid`

Congrats! The metadata file was valid. Results of the split operation are in the `items` array. Not every item will itself be valid. 
- [x] First show the interface where a prefix can be specified for a "Look up in Access Platform" button that calls `slug.resolveMany` with each id prepended by the prefix if it's there. **Question**: do we want to do a separate lookup for preservation, or do we also want to do the preservation lookup when we do this one? Separating the two lookups might be an alternative to the `[ ] Access [ ] Preservation` checkboxes below. **Question**: what are we using to do the preservation lookup? `dipstaging`? Note - only pass in valid items
- [x] List split items in a table, maintaining the order of every item is something we'll want to keep throughout. This sets up a grid that should show every possible thing that will be updated by this DMD task.  For each item, show a row with the following columns:
  - id (from the item's `id` field)
  - label (from the item's `label` field)
  - valid: 
     - For valid items, a preview button that opens up a modal containing the results of `dmdtask.fetchResult` with type `json`. May elaborate on this later as we get feedback.
     - For invalid items, the error found in `splitResult.message`. Nothing further can be done with them.
  - access: blank before slugs are looked up; shows the result of the lookup for this id
  - preservation: same but with the preservation lookup result
- [x] Show, somewhere, `Access, Preservation` checkboxes that determine where updates can go. At least one must be selected for updates to run. (? - **Enhancement**: Add an `Update` checkbox to each lookup result. Each is selected by default.) 
- [x] For now, everything that is updatable will be updated when a big `Update Descriptive Metadata Records` button is pressed.
- [x] When the update is triggered, the front-end will marshal a series of `dmdtask.storeAccess` and `dmdtask.storePreservation` requests corresponding to the successful lookups. A progress indicator of some kind should display. The user shouldn't be able to mess with anything else.
- [x] Once the update is complete, maybe show something that says it's done. 
- [x] The user can repeat updates whenever they like.

Cleanup
- [x] done

Documentation
- [x] done


brittnylapierre:
Hey @RussellMcOrmond  @SaschaAdler I wasnt able to use the library above in my in-browser code. I did find this solution for this specific error:
`Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.`
printed by btoa for 'MasterFile_Serials.mrc': https://crkn.sharepoint.com/:f:/r/sites/platform/Shared%20Documents/Access%20Platform/Sample%20Metadata?csf=1&web=1&e=KOGc6A 

`b64EncodedMetadataFileText = btoa(
          unescape(encodeURIComponent(metadataFileText))
        );
`

From stack overflow:  https://stackoverflow.com/questions/23223718/failed-to-execute-btoa-on-window-the-string-to-be-encoded-contains-characte

It gives the method to decode the string too.

Not sure what the best solution would be.

brittnylapierre:
You can view this interface at: https://access-dev.canadiana.ca/dmd/123

brittnylapierre:
@SaschaAdler I quickly changed the wording to fix https://github.com/crkn-rcdr/Access-Platform/issues/257 before we update prod for monday's demo.

SaschaAdler:
Instead of using `btoa`, we should be using https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL. See the "bonus" in https://ionicframework.com/blog/converting-a-base64-string-to-a-blob-in-javascript/

brittnylapierre:
@SaschaAdler thank you! I'm going to add the comment you have for the improvement to an issue (I did the easy thing for now, copied the solution as-is straight in!) but i'll get it sorted in my next PR.

PR #255, opened 2021-08-17
Import back-end

SaschaAdler:
Setting things up for revamped import from preservation.


SaschaAdler:
A few thoughts here:

1. I've fashioned a new ddoc out of the old one. I haven't tested it yet.
1. Instead of getting a bunch of dipstaging documents (or `LegacyPackage`s as I like to call them), loading them in the browser, and then looking up slugs, we should take advantage of lapin's ability to talk to two databases at the same time.
2. You can see, then, that the `importStatus` function, which filters a LegacyPackage (or a lack of one) into something useful for the front-end, also accepts the results of resolving something into a Noid.
3. A LegacyPackage ImportStatus can have five states:
    1. `not-found`, because it doesn't exist. Useful when looking up user-provided strings
    2. `new`, because it's never been smelted
    3. `processing`, because it's waiting to be smelted
    4. `failed`, because smelting didn't work
    5. `succeeded`, because it did
4. The front-end should point out these five different states somehow.
5. I really want us to nail down which views into dipstaging are actually useful to staff. Obviously they need to use some to start, and so I want to start with prefix-able strings, and `reposManifestDate` interval (which now has the view `byManifestDate` in the new ddoc). I will write some lookups in couch-utils for both. I'd love if the manifestDate lookup uses two [Date inputs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date).

SaschaAdler:
I'm not quite "done" here, for whatever value of done we're going for. I haven't sorted out lapin routes for this stuff yet, for example.

We can merge this now, though, I think. I want to ensure it works with changes made in #234.

PR #269, opened 2021-08-24
Fixing fetch post error, dmd task attachment corruption, and dmd task…

brittnylapierre:
### fetch post request error fix: 

Was to downgrade svelte kit:
dependencies:
- @sveltejs/adapter-node 1.0.0-next.42
+ @sveltejs/adapter-node 1.0.0-next.39
- @sveltejs/kit 1.0.0-next.152
+ @sveltejs/kit 1.0.0-next.146

This solution came from me installing the transformer package I talked about this morning. It's dev dependencies were these versions. I uninstalled the package after our meeting and the code worked as expected. (The versions of these stayed downgraded after uninstalling the transformer package.)

### dmd task attachment corruption fix:
The new convertBlobToBase64 method corrupts the file whereas the old way doesn't. Will work out solution to new method in another PR.

if (metadataFileText) {
  b64EncodedMetadataFileText = btoa(
          unescape(encodeURIComponent(metadataFileText))
  ); //await convertBlobToBase64(file); The new way corrupts the file :-(
}

See md5 check:
Screenshot from 2021-08-24 15-39-38

Additional check, when I use the couch web app to upload the file the attachments look the same:
Screenshot from 2021-08-24 15-40-00

### dmd task attachment content type fix:
Ended up having to switch to this.db.attachment.insert. to get it to work I needed to remove the " that are added to the etag for some reason.

try {
  await this.db.attachment.insert(
    document,
    attachmentName,
    attachment,
    contentType,
    {
      rev: headers.etag.toString().replace(/"/g, ""),
    }
  );
} catch (e) {
  throw createHttpError(e.statusCode, e.error);
}

### DIFF
--- a/packages/couch-utils/src/DatabaseHandler.ts
+++ b/packages/couch-utils/src/DatabaseHandler.ts
@@ -470,17 +470,15 @@ export class DatabaseHandler<T extends Document> {
     }
 
     try {
-      // Apparently this.db.attachment.insert gets very confused by the `rev` qs param
-      await this.client.relax({
-        db: this.name,
-        method: "put",
-        path: `${document}/${attachmentName}`,
-        headers: {
-          "Content-Type": contentType,
-          "If-Match": headers.etag,
-        },
-        body: attachment,
-      });
+      await this.db.attachment.insert(
+        document,
+        attachmentName,
+        attachment,
+        contentType,
+        {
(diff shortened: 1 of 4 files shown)


RussellMcOrmond (approved):
I am fine with a merge, and I will re-open the bug if there is a problem when I test in production and on my own desktop.

Over to @DivyaKrishnan26

PR #270, opened 2021-08-25
Resolving the error with adding members during demo

DivyaKrishnan26:
Nothing fancy yet but a working basic feature. Fancy stuff with validations are to follow.


RussellMcOrmond (approved):
Ahh, that thing that looks like it was intended to be an array of objects that ended up being an array of [key, value] arrays.

I won't be surprised if the originating function gets fixed up later.

brittnylapierre (approved):
Looks good thanks @DivyaKrishnan26!

PR #272, opened 2021-08-27
New collection and manifest lapin routes & access handler methods

brittnylapierre:
### `NewManifest` and `NewCollection` zod schemas and types.
- [x] `NewManifest` should have the following `Manifest` properties: `slug`, `label`, `behavior`, `viewingDirection`, `canvases`. Maybe also `summary`, although we aren't using it much yet.
- [x] `NewCollection` should have the following `Collection` properties: `slug`, `label`, `behavior`, `members`. See above re: `summary`.

### `createCollection` and `createManifest` `AccessHandler` methods 
- [x] should accept a Noid, a User, and the corresponding NewManifest/NewCollection, cause the object to be created in the database, fetch it, and return it. 
- [x] To create the new object, either use the `insert` method I sneaked back into DatabaseHandler, or figure out a good way to do it with an update function. 
- [x] You should ensure the New* object is parsed before sending it anywhere (in a manner similar to how `editManifest` and `editCollection` work). 
- [x] In fact, we probably need to do the same sort of checks in these methods as we do in the `edit` methods; it's possible that we can refactor those checks out. ~To test these methods out, we might need to incorporate the dev noid minter into the couch-utils test context.~ (I added valid noid string instead for now)

###  The lapin `manifest.new` and `collection.new` routes should:
- [x]  Mint a noid for the new object
- [x]  Use the `create` method from above to create a record for it
- [x]  ​return the noid


RussellMcOrmond (approved):
I see what looks to me to be test data, but not only within tests (test manifest and collection creation).  Not quite sure if this is something that is intended to be complete, or only to show UI?

I have no concerns about merging at earlier stages, and being able to demonstrate process to other teams.  I'll do more extensive testing when the functionality is complete (Changes in web interface impacting databases).

Thanks.

brittnylapierre:
> I see what looks to me to be test data, but not only within tests (test manifest and collection creation). Not quite sure if this is something that is intended to be complete, or only to show UI?
> 
> I have no concerns about merging at earlier stages, and being able to demonstrate process to other teams. I'll do more extensive testing when the functionality is complete (Changes in web interface impacting databases).
> 
> Thanks.


thanks Russel! Yeah the front-end is to be worked on and completed. I wanted to get the back-end ready for @DivyaKrishnan26 since the feature she's working on depended on it.

brittnylapierre:
I re-based the branch for main and built prod and it all looks good, so I will merge this into main and continue the front end work in: https://github.com/crkn-rcdr/Access-Platform/pull/273

Issue #210, opened 2021-08-05
Create back-end code for saving new manifests and collections


brittnylapierre:
The solution there is to leave them as partial and then use NewManifest and NewCollection for the non-partial versions. 

The new routes would then accept e.g. a NewManifest object, mint a noid for it, and then call an AccessHandler method that calls an update function that only succeeds if the noid isn't in use (it shouldn't be, but who knows). 

The update function will also need to know the type of the object; @SaschaAdler  can help with its implementation. A note there is that while slug is technically optional, the new object interfaces should require one (and it should be unique).

brittnylapierre:
@SaschaAdler If you have time, it would probably be good to write it up some more implementation details for me on this before you go, since I want to work on this next :) 

What's the general flow for this, I know you've described it to me before but I forget :') It doesn't depend on any packaging of the noid service into access platform right?

Thanks!

SaschaAdler:
1. We'll need `NewManifest` and `NewCollection` zod schemas and types.
    - `NewManifest` should have the following `Manifest` properties: `slug`, `label`, `behavior`, `viewingDirection`, `canvases`. Maybe also `summary`, although we aren't using it much yet.
    - `NewCollection` should have the following `Collection` properties: `slug`, `label`, `behavior`, `members`. See above re: `summary`.
2. `createCollection` and `createManifest` `AccessHandler` methods should accept a Noid, a User, and the corresponding NewManifest/NewCollection, cause the object to be created in the database, fetch it, and return it. To create the new object, either use the `insert` method I sneaked back into DatabaseHandler, or figure out a good way to do it with an update function. You should ensure the New* object is parsed before sending it anywhere (in a manner similar to how `editManifest` and `editCollection` work). In fact, we probably need to do the same sort of checks in these methods as we do in the `edit` methods; it's possible that we can refactor those checks out. To test these methods out, we might need to incorporate the dev noid minter into the couch-utils test context.
3. The lapin `manifest.new` and `collection.new` routes should
    1. Mint a noid for the new object
    2. Use the `create` method from above to create a record for it
    3. Return the noid
4. The admin tool for creating a new object should use the same interface as the editor tools. The route should be something like `/object/new`. When the new object is created, the admin tools should redirect to `/object/$NOID`.

PR #273, opened 2021-08-27
Creating new objects in the admin tools interface

brittnylapierre:
### Requirements
- [x] The admin tool for creating a new object should use the same interface as the editor tools. 
- [x] The route should be something like /object/new. 
- [x] When the new object is created, the admin tools should redirect to /object/$NOID.
- [x] Fix issue of editing manifests being broken (add back .partial() to EditabileManifest)


RussellMcOrmond (approved):
My testing involved checking out the branch and then looking at the JSON that ends up in the database.  I'm leaving any code review to Sascha.

I created a new collection -- seems it assumed "behavior": "unordered".  Is this intended to be a question at the creation stage, or is the intention to allow the type of a collection to be changed later as part of the collection editor?


Manifests look good to me.

I saw no mention of updateInternalMeta.  Is that intended to only be triggered on public/unpublish?

brittnylapierre:
Hey @RussellMcOrmond I'm not sure what the default value for collection behaviour should be, I'll edit it to whatever you think it should be :smile: 

For the updateInternalMeta, I'll make a note of this to ask Sascha when he gets back. I implemented the steps that were given in: https://github.com/crkn-rcdr/Access-Platform/issues/210

Sascha wrote the publish/unpublish requests for us a while back!

RussellMcOrmond:
The default you chose is good -- I just wanted to check that the intention for later was to allow the ability to change in the editor.

Thanks.

PR #281, opened 2021-09-08
Memberlist interface

DivyaKrishnan26:
I have my  questions marked in the commit 

8f41eec.


RussellMcOrmond (approved):
Your main question seems to be: when will this be used.

When there is a textbox of ID's being typed in, rather than only doing a slug lookup, you would do one of these lookups.  This allows you to not only get back the noid (when one exists), but also get back an informative reason why a given slug shouldn't be added.

I don't have a comment on the "resolve the slugs before adding", as I'm of the belief that it is noids and not slugs that should be added (with the noids found from the checkAdditions functionality you are writing).  Whether you want the check to be done in both places (a lookup to help with the visual interface, as well as an additional check when doing the add) is an implementation detail.


Separate: guessing from glancing at the code, the difference between "resolveMany" and "lookupMany" seems to be whether the slug is also returned?  I'm curious why it was decided to have both?

packages/couch-utils/src/handlers/access.ts
```diff
@@ -27,6 +27,22 @@ const AccessDatabaseObject = z.union([Alias, Manifest, Collection]);
...
+  | {
+      slug: Slug;
+      canAdd: true | false;
+      reason: AddMemberError;
+    }
+  | {
+      slug: Slug;
+      canAdd: true | false;
+      id: Noid;
+    };
```
SaschaAdler, line 45: `canAdd` will only ever be `false` in the first case, and `true` in the second.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -268,4 +284,51 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
+    const response: AddMemberRecord = {
+      id: id,
+      slug: "",
+      canAdd: true || false,
+    };
+    const negResponse: AddMemberRecord = {
+      slug: "",
+      canAdd: true || false,
+      reason: "not-found" || "is-self" || "already-member",
+    };
```
SaschaAdler, line 305: I'm not sure why you're initializing these here, especially since something like `true || false` evaluates to `true`.

You'll need to initialize the array that this method should return instead.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -268,4 +284,51 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
     const manifest = await this.get(args.id);
     return Manifest.parse(manifest);
   }
+  async checkAdditions(
+    id: Noid,
+    args: { slug: Slug[] }
+  ): Promise<AddMemberRecord> {
```
SaschaAdler, line 290: This should return a type of `Promise<AddMemberRecord[]>` instead of `Promise<AddMemberRecord>. We're going to return one `AddMemberRecord` for each slug.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -268,4 +284,51 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
+        if (members.id === foundSlug?.id) {
+          negResponse.canAdd = false;
+          negResponse.reason = "already-member";
+        }
+      }
+
+      if (!resolution.found) {
+        negResponse.canAdd = false;
+        negResponse.reason = "not-found";
+      } else if (foundSlug?.id === data.slug) {
```
SaschaAdler, line 327: You want to compare `data.slug` with `slugs` here (and I'd probably have called `slugs` `slug`).

SaschaAdler (changes requested):
See the individual comments.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -287,22 +287,13 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
   async checkAdditions(
     id: Noid,
     args: { slug: Slug[] }
-  ): Promise<AddMemberRecord> {
+  ): Promise<AddMemberRecord[]> {
     console.log("Entry Into checkAdditions", id);
 
     const data = EditableCollection.parse(id);
+    let showMemberRecord: AddMemberRecord = [];
```
DivyaKrishnan26, line 294: Initialising the Array that should be returned.

packages/couch-utils/src/handlers/access.ts
```diff
@@ -316,19 +307,19 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
       console.log("Members", currentMembers);
       for (let members of currentMembers) {
         if (members.id === foundSlug?.id) {
-          negResponse.canAdd = false;
-          negResponse.reason = "already-member";
+          showMemberRecord.push(slugs, false, "already-member");
```
DivyaKrishnan26, line 310: Adding the response value to the array - This always expects to resolve to be for the "true" case scenario

packages/couch-utils/src/handlers/access.ts
```diff
@@ -316,19 +307,19 @@ export class AccessHandler extends DatabaseHandler<AccessDatabaseObject> {
...
+        showMemberRecord.push(slugs, false, "not-found");
+       
+      } else if (data.slug === slugs) {
+        showMemberRecord.push(slugs, false, "is-self");
+      } else {
+        showMemberRecord.push(slugs, true, foundSlug?.id);
       }
     }
-    return negResponse;
+    return showMemberRecord;
```
DivyaKrishnan26, line 323: Since the Array always expects to evaluate "True"case,  so, it expects the reason to be set.
Is the way that I am setting the values with the push right?

RussellMcOrmond:
First a suggestion for testing -- part of your test can be to remove from collections before re-adding, and then all of the existing fixtures will be useful to you to add/remove/etc.

brittnylapierre (changes requested):
Looks good! I'm not sure how to do this like Sascha does lolol

I only saw a three quick things in CollectionContentEditor.svelte after checking out the branch and adding members that I think might be helpful look at. They're all UI based though.

1) At the beginning of resolveMembers, the notFoundSlug and validMember slug arrays should be cleared so that the previous results aren't listed with the current results

2) I needed to add
    notFoundSlug = notFoundSlug;
    validMember = validMember;

to the end of resolveMembers to get the results in the table to show in the UI for my build. I run into this with svelte where it only notices arrays are changes if I assign them again.

3) In the results table, I think using the toString() to list all the items will work, since otherwise it's a bit more work to stack the td's of the same index of the two arrays beside each other with each index being it's own tr row.

Something like this is easier for now:
          `<tr>
            <td class:success={validMember?.length}>
              {validMember?.toString()}
            </td>
            <td class:not-success={notFoundSlug?.length}>
                  {notFoundSlug?.toString()}
             </td>
          </tr>`

I think before the results were being generated to something like:
`<tr>
   <td>valid 1</td><td>valid 2</td><td>invalid 1</td><td>invalid 2</td>
</tr>`

RussellMcOrmond:
So, I just tried to follow my own advise in dev mode.

* I looked up oocihm.8_06941 and got to https://access-dev.canadiana.ca/object/69429/s0vq2s46j98h
* I removed the first member
* When I hit "save" I got an error in the response.

```
[{"id":null,"error":{"message":"[\n  {\n    \"code\": \"invalid_type\",\n    \"expected\": \"array\",\n    \"received\": \"object\",\n    \"path\": [\n      \"data\",\n      \"members\"\n    ],\n    \"message\": \"Expected array, received object\"\n  }\n]","code":-32600,"data":{"code":"BAD_REQUEST","stack":"TRPCError: [\n  {\n    \"code\": \"invalid_type\",\n    \"expected\": \"array\",\n    \"received\": \"object\",\n    \"path\": [\n      \"data\",\n      \"members\"\n    ],\n    \"message\": \"Expected array, received object\"\n  }\n]\n    at ProcedureWithInput.parseInput (/repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/router-25f07e2e.cjs.dev.js:44:13)\n    at ProcedureWithInput.call (/repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/router-25f07e2e.cjs.dev.js:69:24)\n    at Router.invoke (/repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/router-25f07e2e.cjs.dev.js:299:22)\n    at Object.mutation (/repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/router-25f07e2e.cjs.dev.js:318:21)\n    at Object.callProcedure (/repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/callProcedure-ae40050c.cjs.dev.js:63:19)\n    at /repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/index-e4015f05.cjs.dev.js:239:44\n    at Array.map (<anonymous>)\n    at Object.requestHandler (/repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/index-e4015f05.cjs.dev.js:235:45)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async /repo/node_modules/.pnpm/@trpc+server@8.4.2/node_modules/@trpc/server/dist/trpc-server.cjs.dev.js:63:5","path":"collection.edit"}}}]
```

I know removing isn't part of the work for this PR, but I believe one of the obvious tests for this PR isn't available yet.

RussellMcOrmond (approved):
I did a lookup with "No prefix" of the following

```
oocihm.8_06941_1
oocihm.8_06941_1d
oocihm.8_06941
oocihm.45116
```

I got what I expected from the output, and I could add.  

But as mentioned, I couldn't save.  I approve this PR, but suggest a new issue is opened to determine what is wrong with save.  Would be nice to confirm that other developers are having the same problem.

brittnylapierre (approved):
Updated my local branch to give it a go and it looks good! Lol sorry @DivyaKrishnan26  @RussellMcOrmond I do get the same error on save to confirm what you saw

PR #285, opened 2021-09-10
Setting up the lapin route for dipstaging

DivyaKrishnan26:
c10aab9- Commit has the route for retrieving the records from the dipstaging database.

### DIFF
--- a/services/admin/src/lib/components/dipstaging/Dipstaging.svelte
+++ b/services/admin/src/lib/components/dipstaging/Dipstaging.svelte
@@ -1,0 +1,44 @@
+<!--
+@component
+### Overview
+The resolver component allows the user to enter a slug, and then a request is sent to the backend to vheck if an object exists with that slug. Error states are shown for invalid slugs or otherwise.
+### Properties
+|    |    |    |
+| -- | -- | -- |
+| slug : string                      | optional | Slug being resolved. |
+| noid: string | null | undefined    | optional | Noid that the slug resolves to, or null. |
+| hideInitial : boolean              | optional | Whether to hide the display when the current slug is the same as the initial slug provided. |
+### Usage
+```  
+  <Resolver
+    bind:slug={object["slug"]}
+    noid={object["id"]}
+    on:available={(event) => {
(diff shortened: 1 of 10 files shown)


services/admin/src/lib/components/dipstaging/Dipstaging.svelte
```diff
@@ -0,0 +1,44 @@
...
+   * @type {Session} The session store that contains the module for sending requests to lapin.
+   */
+  let prefix = "";
+  const { session } = getStores<Session>();
+  async function showDipstagingRecords() {
+    /*  let slugs = input.split(/[,|\s]/);
+    if (prefix.length > 0) {
+      slugs = slugs.map((slug) => prefix + slug);
+    } */
+    const response = await $session.lapin.query("dipstaging.listFromKeys",);
```
DivyaKrishnan26, line 38: The **dipstaging.listFromKeys** states it doesnt exist, should the route be imported anywhere apart from the router.ts file?

brittnylapierre:
@SaschaAdler @DivyaKrishnan26 @RussellMcOrmond I fixed the dipstaging failing test here.

I finished up the route tests in admin too! Please wait until the tests complete running, then navigate to:

https://access-dev.canadiana.ca/import/keys/test
Screenshot from 2021-09-21 16-40-25

and

https://access-dev.canadiana.ca/import/dates/2012-01-01,2021-09-09
Screenshot from 2021-09-21 16-41-12

If/when approved I'll start my own PR and branch (es) as needed for the rest of the smelter stuff.

RussellMcOrmond (approved):
I'm fine with the work so far.

However, I believe the interface should allow for more generic access to views.  There are currently 5 relevant views to be looking at, and it is likely best if this is a parameter rather than there being a separate route for all 5.  I only glanced at code, so don't know if this is how things already work.

https://github.com/crkn-rcdr/Access-Platform/tree/main/services/couchdb/dipstaging/design/access/views

If you would like a demo of https://admin.canadiana.ca/archivalmanifest , let me know.

PR #287, opened 2021-09-11
Admin app improvements based on initial feedback

brittnylapierre:
- [x] Improved styling for side menu (not sure if good enough to be final)
- [x] Drop down component closes when item clicked
- [x] Slug and label shown in add manifest menu


SaschaAdler:
I don't see anything immediately objectionable, code-wise, here. We might want to move away from magic width numbers in media queries at some point.

SaschaAdler (approved):

Issue #302, opened 2021-09-16
Periodic poll for "Descriptive Metadata Updater"


RussellMcOrmond:
After an upload the next page says "Please wait" with a cool animated image suggesting activity.

What I note from watching the developer tools is that it isn't periodically polling for updates.

Suggestion:

* Launch a event to poll every once in a while (30 seconds?)
* Add a third date for "Last checked" that updates with the poll time.

Thanks.

PR #303, opened 2021-09-17
Fixes for admin app issues/enhancements

brittnylapierre:
- [x] #298 
- [x] #299
- [x] #302
- [x] Documentation

### DIFF
--- a/services/admin/src/lib/components/dmd/DmdSuccessView.svelte
+++ b/services/admin/src/lib/components/dmd/DmdSuccessView.svelte
@@ -169,7 +169,11 @@ This component shows the view for a dmd task that had its metadata successfully 
     {:else}
       <NotificationBar
         status="fail"
-        message={`Every item's metadata is invalid. To get an idea of what went wrong for each item, press their preview button. You can use this information to make changes to ${
+        message={`${
+          dmdTask?.items?.length
+            ? "Every item's metadata is invalid. To get an idea of what went wrong for each item, press their preview button. You can use this information to m"
+            : "No items found. M"
+        }ake changes to ${
           dmdTask?.fileName?.length ? dmdTask?.fileName : "the file"
         }, then try again by uploading the corrected file.`}
       />
--- a/services/admin/src/lib/components/dmd/DmdSuccessView.svelte
+++ b/services/admin/src/lib/components/dmd/DmdSuccessView.svelte
@@ -55,6 +55,11 @@ This component shows the view for a dmd task that had its metadata successfully 
   let showAllInvalidError: boolean = false;
 
   /**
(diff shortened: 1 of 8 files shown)


services/admin/src/lib/components/dmd/DmdSuccessView.svelte
```diff
@@ -169,7 +169,11 @@ This component shows the view for a dmd task that had its metadata successfully
     {:else}
       <NotificationBar
         status="fail"
-        message={`Every item's metadata is invalid. To get an idea of what went wrong for each item, press their preview button. You can use this information to make changes to ${
+        message={`${
+          dmdTask?.items?.length
+            ? "Every item's metadata is invalid. To get an idea of what went wrong for each item, press their preview button. You can use this information to m"
```
SaschaAdler, line 174: I think you cut off this error string from what it was.

SaschaAdler (changes requested):
Looks fine except for my one comment.

services/admin/src/lib/components/dmd/DmdSuccessView.svelte
brittnylapierre: Ah I kind of added a bit of inline logic, here's the full line:

`
${ dmdTask?.items?.length ? "Every item's metadata is invalid. To get an idea of what went wrong for each item, press their preview button. You can use this information to m" : "No items found. M" }ake changes to ${ dmdTask?.fileName?.length ? dmdTask?.fileName : "the file" , then try again by uploading the corrected file.`}
`

So the message depends on the items list length. I can make this a variable and use the $: codeblock instead to make it more clear?

SaschaAdler: Oh, I see what you did. I'm not sure this is ideal, and I wouldn't recommend doing things this way for anything we wanted to provide in multiple languages, but it's fine for now.

SaschaAdler (approved):

services/admin/src/lib/components/dmd/DmdSuccessView.svelte
brittnylapierre: Ohh okay! I've never done multi language support on a website I'm excited to get there, thanks Sascha!

brittnylapierre:
Hey @RussellMcOrmond since it was quick i added the polling here too, let me know if it looks good to you! I'll merge and trigger a new build if so

RussellMcOrmond (approved):
Given these are new things and not something that could break something live, I'm happy to review after a deploy to production.  I assume you ran both `pnpm dev` against dev data and `pnpm prod` against production data.

brittnylapierre:
Tested it with pnpm run prod! I'll merge and build for production now

PR #311, opened 2021-09-20
Adding Collection Behaviour to the Editor

brittnylapierre:
You may need to run pnpm install, there's a new virtual scrolling library added

The behaviour is added to the general info tab, for collections only.

The only options are: unordered, multi-part.

If the collection is unordered, the user can only delete items, not re-order them in the manage members tab.

### DIFF
--- a/services/admin/src/lib/components/access-objects/InfoEditor.svelte
+++ b/services/admin/src/lib/components/access-objects/InfoEditor.svelte
@@ -81,17 +81,20 @@ This component displays the non content properties for an access editorObject an
         }}
       /><br /><br />
 
+      {#if isCollection(editorObject)}
+        <label for="behavior">Behavior</label><br />
+        <select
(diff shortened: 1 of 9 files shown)


DivyaKrishnan26 (approved):
A small thing, I am not sure if we want to have the Behaviour part on the bottom or if we want to have that on top after the slug.  And another small thing do we want to correct the Spelling for the "Behavior" to "Behaviour"?

So, basically this toggle lets us know if the Collection is Multi-part or unordered right? Can we have a multi-part collection be changed to an unordered collection and save it?- this is something i would like to know.

RussellMcOrmond (approved):
I have no opinion on the placement of the buttons.  I've done my own test in the dev environment, and seems to do the trick.

DivyaKrishnan26:
Can we merge this then? So that I can create a branch of the latest base?

PR #312, opened 2021-09-20
Fix broken staff objects

### LINKED ISSUE #293: `staff` field with `"by": {}` with access/updates/publish and unpublish
Reading https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/access/design/access/updates/publish.js and https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/prelude.js 

it suggests that not sending a body is valid.  

Except, when I try that, I get a 400 error.  I suspect it is because the parse of the JSON isn't within a try{}

I've been sending {} which is valid JSON, but the fact it is valid json means the current test in updateObject() thinks there is a user, and puts the {} into the `by` field.  That may be causing issues in other places which assume that field is fully filled in if it exists.  It may be the cause of a Zod error I've seen when trying to edit a manifest.

### LINKED ISSUE #294: Zod error on editor load
https://access.canadiana.ca/object/69429/m0m901z46797
Screenshot from 2021-09-20 10-40-05

SaschaAdler:
Resolves #293.
Resolves #294.


SaschaAdler:
@RussellMcOrmond I've run the code already, and there should no longer be any broken staff objects. I'll merge this now, but let you confirm before the issues are closed.

Issue #318, opened 2021-09-22
Date parsing not consistent.


RussellMcOrmond:
I note from looking at https://access.canadiana.ca/object/69429/m0fj29883036

"Status: published on 31/12/1969, 19:33:40"

Dates can be in two formats, and this parsing should probably happen within Lapin to make consistent.

Related: I'm regularly needing to convert the seconds-since to a string when using Fauxton for checking the state of data.  Having the human readable version of the date be what is in the database would be ideal, even if the numeric version is used in all the applications.

SaschaAdler:
I agree. I propose we use the ISO 8601 date specification. All of it, for maximal interoperability and minimal "having to think about it". If anything has a hard time with fractional seconds (which, btw, Solr doesn't, at least as of the link you shared) or non-UTC time zones, we take care of it then.

If we restrict ourselves unnecessarily, we then have to focus on ES, CouchDB, Perl, Solr, MySQL, Blazegraph, Fortran, or whatever.

PR #320, opened 2021-09-22
Dipstaging lookup UI

brittnylapierre:


RussellMcOrmond:
When I first ran as "dev" there was some odd warnings when looking at https://access-dev.canadiana.ca/smelter/queue .  My guess is that the issue is that the couchDB view hadn't been calculated yet and the result wasn't what was expected.

When I run as "prod" to look at the views I get "Internal server error" for Queue, just a list (No details, and' can't do anything) for "New Dips",  A list (no details, can't do anything) for "Never Smelted", and nothing under status (Expected as there are currently no errors).

I tried a lookup -- I'm not sure if the "comma separated" is a requirement of the tool.  I pasted in a list (newlines) which is what staff will want to do, and didn't get anything from the lookup.

brittnylapierre:
@RussellMcOrmond @SaschaAdler What's the difference between new dips view and dipstaging docs with smelt.status = success? Thanks

RussellMcOrmond:
Views are here: https://github.com/crkn-rcdr/Access-Platform/tree/main/services/couchdb/dipstaging/design/access/views

* byManifestDate.js  -- Generic view that shows every document with a reposManifestDate (which should be everything), and emits a date array
* neverSmelted.js  -- Has a reposManifestDate , but doesn't have a "smelt" field -- meaning, it hasn't been asked to be smelted yet.  Emits that common date array for reposManifestDate
* newDip.js  -- compares the reposManifestDate and the doc.smelt.processdate and if they both exist and the ManifestDate is more recent then emit the familiar date array.
  * This is new data in the Preservation platform since the date Smelter was run.
  * The Preservation Platform uses [OAIS lterminology](https://en.wikipedia.org/wiki/Open_Archival_Information_System#The_OAIS_environment_and_information_model) such as SIP, AIP and DIP.
  * The Access Platform uses [IIIF terminology](https://iiif.io/api/presentation/3.0/#12-terminology). We currently support collections, manifests, and a specific single-image type of canvas.
* smeltQueue.js --  entries that are in the queue to be processed by Smelter (has a "smelt" field, and the request date is more recent than the process date -- which might be 0).  This emits a date array, but for the requestDate so they are sorted correctly
* smeltStatus.js -- entries that have a smelt field, where there is a message, and the processDate is more recent than the requestDate (meaning, it is done).  It emits a boolean indicating if the processing succeeded or not and that date array.  This one has a value for the emit, which is the message.


Hope that clarifies.  

newDip is all about the SIP update problem discussed in the "platform team" forum in relation to our current custom preservation platform (to be replaced by Archivematica).

smeltStatus is about seeing results from the Smelting process that extracts data from the Preservation platform and inserts into the Access platform.  The only results staff care about are those that weren't a silent success -- most will just work.

Smelter is the gateway between Preservation and Access, and that is why you see both sets of terminology used in this one tool. 


While AIP Identifiers in preservation may look the same as Slug identifiers in access, the ability for these to be different is a key motivation for the PASplit project.  There will be more changing of the guessed slugs in the future, as the guess from the Preservation AIP ID won't be as helpful. AIP IDs and Slug ID's won't be the same as Archivematica deliberately includes an opaque UUID within the AIP ID to help discourage people from thinking these identifiers have transparent meaning.

brittnylapierre:
Thanks so much Russell that helps a lot! :smile:

brittnylapierre:
See the 3 todos left in the description - but I'd like to get some feedback on the UI so far

RussellMcOrmond (changes requested):
Testing with `pnpm prod` as I wanted to look at some of the real data, as there is quite a bit in the Smelting queue at the moment.

In production we have a sample for "Smelter Results" which was a success, but where there were many warnings. There are improvements to Smelter I can make, but this is helpful for the UI.

The messages should always be put into a &lt;pre&gt;&lt;/pre&gt; block, as they use newlines to separate messages.

I got a "zod" error on the "Smelting Queue".

I tried the "Look-up by Slug" using some of the [ID's in Beth's list](https://crkn.sharepoint.com/:w:/s/HeritageTech/Eeo0yxJPNlRNu5exscA8baYBTBS_L8K6aCUFSUT_yKiRWg?e=4UKtO1), and the lookup didn't seem to work.

```
N_00273_190508 

N_00273_190509 

N_00273_190708 

N_00273_190808 

N_00273_190908 

N_00273_191008 

N_00273_191109 

N_00273_191210 

N_00273_191312 

N_00273_191602 
```

brittnylapierre:
Okay sounds good I'll change it back! Lolol that makes the speed make more sense :')

RussellMcOrmond:
> Okay sounds good I'll change it back! Lolol that makes the speed make more sense :')

You may want to pause and give that a thought.

When I think about things, I believe you want to show all results that have a message, true or false.  Maybe `update.succeeded` should be part of the value rather than part of the key?  Just an idea -- see is that makes more sense to you.

```
emit(dateAsArray(processDate), [update.succeeded,update.message]);
```
At the moment there isn't a "just show me the ones that failed" and "show me the ones that succeeded but had a message".  Having the success be part of what you display, regardless of the order, may be what you want?

Also -- if you are going to grab the whole document anyway, you can emit `null` for the value of the emit.  You don't need to emit anything, and that actually makes the view more disk efficient.  What document was emit from is automatically included.

RussellMcOrmond (approved):
At this point the changes being discussed are cosmetic, and not about basic functionality.  Merging, deploying, and having other people be able to make comments would be desirable.

RussellMcOrmond (changes requested):
I just did a "Run Smelter on Selected Packages" after doing a lookup, and the UI suggests that Smelter is running on all the packages and not only the ones that were checked.

The tool showed the checkbox not selected for anything where Smelter had already run successfully, as well as for any that Smelter was currently running.  This is as it should be.

I looked in the actual queue on the database, and the database queue wasn't increased for those items.  Is it possible the display has a cache of something you updated but didn't need to be updated in the database?



```
N_00273_190508 

N_00273_190509 

N_00273_190708 

N_00273_190808 

N_00273_190908 

N_00273_191008 

N_00273_191109 

N_00273_191210 

N_00273_191312 

N_00273_191602 

N_00311_191507 

N_00311_191806 

N_00319_189909 

N_00319_190008 

N_00319_190108 

N_00319_190208 

N_00319_190308 

N_00319_190408 

N_00531_185609 

N_00531_185803 

N_00531_186408 

N_00531_187104 

N_00531_187710 

 

 
```

RussellMcOrmond:
I should clarify the two different issues I detected with "Run Smelter on Selected Packages"

1. The tool didn't seem to look at the checkmarks, and tried to initiate smelter on all the packages on the view rather than only the selected ones.
2. The update function is using "seconds since" as a date format, when Smelter is using the database format as designed which uses ISO date strings.  See: https://github.com/crkn-rcdr/Access-Platform/tree/main/deprecated/Databases/dipstaging

My suggestion on the second is to add functions to the prelude which can be used to cast a date between seconds since and the ISO date format we are using. Most of our databases use the ISO date format, and having a single common format in the databases makes exchanging data between databases much easier.


Example generating:

https://github.com/crkn-rcdr/Access-Platform/blob/89a0d0e4627af34466e217eaccaee932f3cbe7e0/deprecated/Databases/dipstaging/design/sync/updates/basic.js#L3-L5

RussellMcOrmond (approved):
I did my "Look-up by Slug" and was able to select a single checkbox and that worked (in production it is the old update document until we merge and push that).

I notice on "Smelting Queue" that the dates show 1970, but that will relate to #318.

I recommend this be merged and deployed, with other enhancements as separate PR's.

PR #321, opened 2021-09-22
Getting the members context backend code

DivyaKrishnan26:


packages/lapin-router/src/routes/collection.ts
```diff
@@ -58,4 +62,18 @@ export const collectionRouter = createRouter()
...
     },
+  })
+  .query("viewMembersContext", {
+    input: addMemberscontext.parse,
+    async resolve({ input, ctx }) {
+      try {
+        return await ctx.couch.access.findUniqueArray("id", input.id, [
+          "slug",
+          "label",
+        ]);
```
SaschaAdler, line 73: The thing I wanted to add here was that you should add `as const` after the field array, like so:

```ts
        return await ctx.couch.access.findUniqueArray("id", input.id, [
          "slug",
          "label",
        ] as const);
```

This means that the objects you get back will be aware that they have `slug` and `label` fields.

DivyaKrishnan26: Right, got it. I will change that and push it now.

services/admin/src/lib/components/collections/CollectionContentEditor.svelte
```diff
@@ -53,6 +56,15 @@ Allows the user to modify the member list for a collection.
       setActiveIndex(activeMemberIndex);
     }
   }
+  async function getMemberContext() {
+    for (let index of collection.members) {
+      const resolutions = await $session.lapin.query(
```
DivyaKrishnan26, line 61: I think if we add in the for here its going to be too laborious and costly. but I just added to check if it works but still its a 404 for me.

SaschaAdler: What kind of 404 are you getting? Also, is VSCode giving you an error when you use the query?

DivyaKrishnan26: No, VS code doesn't give me error, but the console gives me a 404 and now this,
```
Uncaught (in promise) TRPCClientError: No "query"-procedure on path "collections.viewMembersContext"
 
```

SaschaAdler: I think it's `collection.viewMembersContext` instead of `collections.viewMembersContext`. VSCode should be giving you an error on the latter.

DivyaKrishnan26: My bad, I changed that now its not a 404 anymore. And the error now makes sense.

DivyaKrishnan26: So, the error now tells me it expects an object and I am passing a string, My question is the noid is a string right? are we not supposed to pass in that?
```
CollectionContentEditor.svelte:16 Uncaught (in promise) TRPCClientError: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "string",
    "path": [],
    "message": "Expected object, received string"
  }
]
    at Function.from (TRPCClientError-7653e11e.esm.js:50)
    at transformRPCResponse (httpRequest-9a687d18.esm.js:33)
    at trpc-client-links-httpBatchLink.esm.js:187
```

services/admin/src/lib/components/collections/CollectionContentEditor.svelte
```diff
@@ -59,7 +59,7 @@ Allows the user to modify the member list for a collection.
   async function getMemberContext() {
     for (let index of collection.members) {
       const resolutions = await $session.lapin.query(
-        "collections.viewMembersContext",
+        "collection.viewMembersContext",
         index.id
```
DivyaKrishnan26, line 63: Its Zod and TRPC showing a type mismatch but then I am not sure. I have changed the input again to take in a noid and supply a string the mismatch still exists, What exactly should be the input and the output here? So that TRPC hit me with the type mismatch?

DivyaKrishnan26: Resolved this in the next commit

services/admin/src/lib/components/collections/CollectionContentEditor.svelte
```diff
@@ -57,7 +57,7 @@ Allows the user to modify the member list for a collection.
       setActiveIndex(activeMemberIndex);
     }
   }
-  let documentSlug: ObjectList = [];
+  let documentSlug: string[] = [];
   async function getMemberContext() {
```
DivyaKrishnan26, line 61: What type should the object be, should I be creating a type similar to the Object list which has slug and the label fields?

services/admin/src/lib/components/collections/CollectionContentEditor.svelte
```diff
@@ -144,6 +144,13 @@ Allows the user to modify the member list for a collection.
                     />
                   {/if}
                 {/each}
+                {#each documentSlug as document}
+                  <textarea
+                    id="label"
+                    name="label"
+                    bind:value={document.slug}
```
DivyaKrishnan26, line 151: This is where I get the error. (apologies the check button I will get that away once the values are retrieved right)
```
CollectionContentEditor.svelte:151 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'slug')
    at Object.mount [as m] (CollectionContentEditor.svelte:151)
    at Object.update [as p] (CollectionContentEditor.svelte:147)
    at update_slot (index.mjs:99)
    at Object.update [as p] (VirtualList.svelte? [sm]:248)
    at update_slot (index.mjs:99)
    at Object.update [as p] (DynamicDragAndDropListItem.svelte? [sm]:42)
    at update (index.mjs:1044)
    at flush (index.mjs:1012
```

DivyaKrishnan26:
Work done in #356

PR #332, opened 2021-10-04
Membership re-do

SaschaAdler:


RussellMcOrmond:
I'd like to get a confirmation on `pageLabels`, which came to mind with noids no longer being optional for an `ObjectList`  (For some reason I thought it was optional because of born digital).

It is listed as deprecated, but is that being left as a future change?  Smelter currently uses it for born digital, which we need to confirm with our changes are still working.

https://github.com/crkn-rcdr/CIHM-Meta/blob/main/lib/CIHM/Meta/Smelter/Process.pm#L174-L184

SaschaAdler:
```ts
    /**
     * Labels for this PDF's pages. Deprecated, as Canvases will be generated
     * for `pdf` Manifests.
     */
    pageLabels: z.array(TextRecord).optional(),
```

`pageLabels` doesn't use `ObjectList`.

RussellMcOrmond:
> Do we care about `seq` for unordered collections?

As far as I can think of, we don't care one way or the other (IE: do whatever makes life easiest to manage).

Separate from database or Lapin API, the user interface may want to display the recently added to the top to make it easier for staff to notice they are there.

services/haproxy/haproxy.cfg
```diff
@@ -41,7 +41,7 @@ frontend public
   bind :443 ssl crt /usr/local/etc/haproxy/certs/canadiana.pem alpn h2,http/1.1
 
   # HTTPS always. Comment if this isn't working for you.
-  http-response set-header Strict-Transport-Security max-age=63072000
+  # http-response set-header Strict-Transport-Security max-age=63072000
```
RussellMcOrmond, line 44: Curious if it would be easier to update the docker-compose with a ports: for couchdb, rather than changing this config.  I worry that this will allow us to miss something HTTPS specific while testing in dev that we need to work in prod.

SaschaAdler: It's my understanding that this file, as-is, doesn't get used for prod, but the specific issue is that couchdb-nano chokes on HTTPS interactions with certs it doesn't understand.

You're welcome to suggest an amendment or point me to how to have this only work for couch; I'm not sure that I have the time to figure out how to do this port-specific.

PR #336, opened 2021-10-06
Lapin test harness

DivyaKrishnan26:
Questions: 
Should the new test environment replicate the couch-utils test, as it should it have its own Kivik, DatabaseHandler? If we are to create in the new server for the test harness.
What would be the ServerScope for the environment?
The Lapin router is the wrapper to translate the the couch-calls right, so the REST endpoints need to be tested? 

(I have tried creating new, but I am not sure if that is what is needed)

### DIFF
--- a/packages/lapin-router/package.json
+++ b/packages/lapin-router/package.json
@@ -69,6 +69,7 @@
     "@crkn-rcdr/couch-utils": "workspace:*",
     "@crkn-rcdr/swift": "workspace:*",
     "@trpc/server": "^8.4.2",
+    "ava-ts": "^0.25.2",
     "deep-object-diff": "^1.1.0",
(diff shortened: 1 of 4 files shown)


SaschaAdler:
Forget about spinning up new copies of test databases, for now.

Can you come up with something that lets us test one of the query routes? Those won't change any data.

DivyaKrishnan26:
> This looks promising so far. What happens when the tests run?

There is still some error that comes, which I am not able to figure out, the error comes, the error is mostly how the referencing is done, its what the typescript complains. And I think resolving that we could run something on to see what the test returns, right now the run complains about the error that comes.

SaschaAdler:
Closing this for now. I'll be revisiting this conceptually shortly.

PR #353, opened 2021-10-19
Upgrade to CouchDB 3.2.0

RussellMcOrmond:
... and allow access to Couch via local OS (and not only via docker network).


RussellMcOrmond:
I don't expect any differences as far as the apps are concerned, but I wanted us to be testing early.  The hope is to upgrade soon.

I'll merge unless there are objections to these two minor changes.

PR #356, opened 2021-10-21
Members context display

DivyaKrishnan26:
The Slug and the Label for Collection members are displayed, The drag and drop works for this. 
My assumptions the logic of the code will change when we work specifically on Pagination part of members

### DIFF
--- a/packages/lapin-router/src/routes/collection.ts
+++ b/packages/lapin-router/src/routes/collection.ts
@@ -22,6 +22,8 @@ const checkAdditions = z.object({
   id: Noid,
   slugArray: z.array(Slug),
 });
+
+const id = z.array(Noid);
 export const collectionRouter = createRouter()
   .mutation("edit", {
     input: EditInput.parse,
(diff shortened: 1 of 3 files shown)


RussellMcOrmond (approved):
I looked at the new toggling, and how "Look_up_slugs" stays available.

I notice that it doesn't allow me to actually look-up the second time, even though the button is there.

Ahh, if I hit "Clear Text" first then it works.  I guess something wasn't reset from the earlier lookup and add.

DivyaKrishnan26:
> I looked at the new toggling, and how "Look_up_slugs" stays available.
> 
> I notice that it doesn't allow me to actually look-up the second time, even though the button is there.
> 
> Ahh, if I hit "Clear Text" first then it works. I guess something wasn't reset from the earlier lookup and add.

Can you test now? I was not able to replicate the issue, Although I am making one small change which I feel its a valid one to do, I would want to disable and enable the dd button only when the element is available to add

RussellMcOrmond:
> Can you test now? I was not able to replicate the issue, Although I am making one small change which I feel its a valid one to do, I would want to disable and enable the dd button only when the element is available to add

I think you forgot to push your commit..

DivyaKrishnan26:
I am merging this PR to main, and have anything changed respected to this along with the pagination works.

Issue #368, opened 2021-10-27
Create microservice to extract OCR data from old preservation platform


RussellMcOrmond:
Smelter currently has some temporary hacks such that legacy OCR data is processed at the same time as images.  This is introducing some complexities when trying to enhance the processing of Canvases.

Example: When Smelter/etc are looking up existing canvases, this will be looking up the image.  #333

If a SIP was updated to add OCR, the OCR will not be copied as part of that process as existing canvases will be found rather than new ones created.

Copying PDF and image data at a later time than METS processing also makes it possible for the OCR XML data to not match the image after a SIP update.

Copying of OCR data should be separated from Smelter earlier, allowing a separate process to "update" OCR from legacy AIPs.  Handling single-page PDFs would be combined with the work for #153.

Tool should detect dummy txtmap https://github.com/crkn-rcdr/Access-Platform/issues/362 as well as blank (only white-space) OCR XML.

RussellMcOrmond:
* [ ] Test for dummy OCR
* [ ] Look up parents of changed canvases and do design/access/_update/forceUpdate to ensure Hammer2 runs.

PR #370, opened 2021-11-03
Changes required to support separating of OCR data processing from Smelter.

RussellMcOrmond:
This is related to https://github.com/crkn-rcdr/Access-Platform/issues/368

* Smelter will look up to see if image canvases already exist using [_view/cihmsource](https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/canvas/design/access/views/cihmsource.js) , and create the canvases otherwise.
* Re-smelting can't be used to add OCR to images that have already been smelted, so a different process will exist that will find the OCR that matches the image, and attach that to the canvas.
* The UI should allow people the option of processing the OCR.  This can use the same tool as smelting, but should be different button as they may happen at different times.

### DIFF
--- a/packages/data/src/import/LegacyPackage.ts
+++ b/packages/data/src/import/LegacyPackage.ts
@@ -76,6 +76,11 @@ export const LegacyPackage = z.object({
   smelt: SmeltProcess.optional(),
 
   /**
+   * Request to extract OCR information from his package.
+   */
+   ocr: SmeltProcess.optional(),
+
+   /**
    * A record of the staff member who caused this record to be updated.
    */
   staff: StaffUpdate.optional(),
--- a/services/couchdb/canvas/design/access/views/cihmsourcepath.js
+++ b/services/couchdb/canvas/design/access/views/cihmsourcepath.js
@@ -3,6 +3,5 @@ module.exports = {
     if (typeof doc.source == "object" && doc.source.from === "cihm") {
       emit(doc.source.path, null);
     }
-  },
-  reduce: "_count",
+  }
 };
--- a/services/couchdb/canvas/design/access/views/copycanvasfiles.js
+++ b/services/couchdb/canvas/design/access/views/copycanvasfiles.js
@@ -1,10 +1,6 @@
 module.exports = {
   map: function (doc) {
-    if (
-      doc.orphan !== true &&
-      (("master" in doc && "path" in doc.master) ||
-        ("ocrPdf" in doc && "path" in doc.ocrPdf))
-    ) {
+    if (doc.orphan !== true && "master" in doc && "path" in doc.master) {
       emit(null, null);
     }
   },
--- a/services/couchdb/dipstaging/design/access/updates/cancelOCR.js
+++ b/services/couchdb/dipstaging/design/access/updates/cancelOCR.js
@@ -3,7 +3,7 @@ module.exports = function (doc, req) {
     successReturn,
     errorReturn,
     extractJSONFromBody,
-    timestamp,
+    updateGenericObject,
   } = require("views/lib/prelude");
 
   if (!doc) {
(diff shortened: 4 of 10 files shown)


RussellMcOrmond:
A "SmeltProcess" appears to be a "ProcessUpdate", but where more of the other fields are allowed to be optional.

Is it worth developers thinking about whether only one of these types is needed?  Having ProcessUpdate be more relaxed might make a number of things easier to process.

RussellMcOrmond:
I started this, but suspect a different developer with more familiarity with the current developer environment can clean it up.  This PR isn't assigned.  I hope to be able to proceed with this work some time next week, after doing the Smelter work.


In https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/access/design/access/updates/unpublish.js it uses an  `updateObject(doc);` from https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/prelude.js#L78-L89

That function is unfortunately specific to the "access" database, although it could be made generic if the modification of the `updateInternalmeta` field was moved into a different named function that called this one. Or just create a separate generic function with similar functionality.  Whatever the devs think makes sense.

This type of prelude would allow `updateObject(doc, user);` to replace the direct changing of doc.user and doc.updated in the updates in dipstaging/access as well, and allow setting user to be optional.  As with those updates for the access database, there are times when smelting and OCR processing will be initiated by automated processes that probably shouldn't be changing the `user` field.

RussellMcOrmond:
If there are no objections, I would't mind merging and letting CouchDB in production build the indexes some time on Friday, so that I can restart the canvas copying over the weekend.

I can make a copy of `updateObject(doc, user);` and use that to allow `user` to be blank, and proceed if that is fine.

Thanks.

RussellMcOrmond:
Noting that https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/access/design/access/lib/prelude.js and https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/dipstaging/design/access/lib/prelude.js are symbolic links.

I've added an  `updateGenericObject()`, and developers can decide how they wish to clean that up later.

RussellMcOrmond:
Some of the tests I used, in case they can/should be added into test suite:


* POST required...

```
curl -v -X POST http://localhost:5984/dipstaging/_design/access/_update/cancelOCR/newDip
```

* JSON body required

```
curl -v -X POST -d '{}' http://localhost:5984/dipstaging/_design/access/_update/cancelOCR/newDip
```

* Can set slug while requesting Smelt

```
curl -v -X POST -d '{"slug": "testSlug"}' http://localhost:5984/dipstaging/_design/access/_update/requestSmelt/newDip
```

* Can request and cancel without changing user or slug

```
curl -v -X POST -d '{}' http://localhost:5984/dipstaging/_design/access/_update/requestSmelt/newDip
curl -v -X POST -d '{}' http://localhost:5984/dipstaging/_design/access/_update/cancelSmelt/newDip
curl -v -X POST -d '{}' http://localhost:5984/dipstaging/_design/access/_update/requestOCR/newDip
curl -v -X POST -d '{}' http://localhost:5984/dipstaging/_design/access/_update/cancelOCR/newDip
``` 

* Can change user during request:

``` 
curl -v -X POST -d '{"user": { "email": "test@test", "name": "russell1" }}' http://localhost:5984/dipstaging/_design/access/_update/requestSmelt/newDip
curl -v -X POST -d '{"user": { "email": "test@test", "name": "russell2" }}' http://localhost:5984/dipstaging/_design/access/_update/cancelSmelt/newDip
curl -v -X POST -d '{"user": { "email": "test@test", "name": "russell3" }}' http://localhost:5984/dipstaging/_design/access/_update/requestOCR/newDip
curl -v -X POST -d '{"user": { "email": "test@test", "name": "russell4" }}' http://localhost:5984/dipstaging/_design/access/_update/cancelOCR/newDip
```

PR #371, opened 2021-11-03
Single operation list updates

SaschaAdler:
This should handle atomic updates to the object that contains the list. When updating collection members, operations should probably also `forceUpdate` each affected member, although I'm not sure when to do this, exactly.

I've only tested it by patching `removeMember` to use it so far (it gets tested through `unassignSlug`).


DivyaKrishnan26 (approved):

brittnylapierre (approved):
Super cool to see, thanks Sascha!

PR #374, opened 2021-11-09
Delete button replaces assign slug button

brittnylapierre:
### Todos:
- [x] change  'Unassign Slug' button to say 'Delete'
- [x] only make delete button visible when unpublished
- [x] check updateInternalmeta state before allowing deletion otherwise show please wait and poll until deletion allowed or 30 mins goes by


DivyaKrishnan26 (approved):

RussellMcOrmond (approved):
Looks good to me.

PR #377, opened 2021-11-16
Membership display / editing

### LINKED ISSUE #94: Retrieve an access object's membership / ancestry
A manifest of collection can be a member of one or more collections.

An accessor method is required for retrieving the collection(s) a given object is a member of. This method can be used recursively to determine the full ancestry of an access object, i.e. figuring out the membership of the collections the object is a member of, etc. etc.

It might be worth memoizing the recursive lookup because the further back you go, the more likely it is you're looking up the same collections over and over.

SaschaAdler:
Closes #94, with the caveat that this isn't going to display the entire ancestry tree all at once, for probable performance nightmare reasons. Users can click links to view the collections this object is a member of, and can also remove the object's membership in a collection.

I'll also fix a bug in which components don't seem to reset properly when traversing between objects. This is a known Svelte issue (feature?) that we figured out how to work around in Sapindale.

### DIFF
--- a/packages/couch-utils/src/handlers/access.ts
+++ b/packages/couch-utils/src/handlers/access.ts
@@ -14,6 +14,9 @@ import {
   NewCollection,
   NewManifest,
   TextRecord,
+  SimpleRecord,
+  Membership,
+  Ancestry,
 } from "@crkn-rcdr/access-data";
 
 import { DatabaseHandler } from "../DatabaseHandler.js";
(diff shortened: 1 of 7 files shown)


SaschaAdler:
The remove button just does its business without any confirmation. We should probably add something to do that but figuring out the reactivity to get that to work is giving me a headache.

SaschaAdler:
I'm happy to merge this as in when everyone is ready.

DivyaKrishnan26 (approved):
We can merge it :)

brittnylapierre (approved):

PR #387, opened 2021-11-24
Make UI compatible with paged API 

brittnylapierre:
- [x] For all simple input fields, add a save button beside the field itself that appears when the field value has changed. Do change comparisons per field, not on the entire object, to make it more performant. 
- [x] Make the Virtual Scrolling component an Infinite Scroll component 
- [x] Make the infinite scroller work with drag and drop 
- [x] Hook up the infinite scroller to the new paging API
- [x] list editing occurs in real time (requests to edit the list are sent as the user interacts with it.) A lock is shown that greys out the interface while any change request is happening, that says something like 'please wait for the change to save...' So that users don't make multiple changes at once to the list. 
- [x] Hook up the remaining operations (add, remove, delete, move through number input, label editing) for canvases
- [x] Hook up the remaining operations (add, remove, delete, move through number input) for members


brittnylapierre:
Opening up for review so I can know what could be better :)

Issue #384, opened 2021-11-18
Update Metadatabus to support "PDF" style Access documents


RussellMcOrmond:
https://numeris-demo.canadiana.ca/view/numeris.RD_2004_FA00_013
Has missing information in http://iris.tor.c7a.ca:5984/_utils/index.html#database/copresentation2/numeris.RD_2004_FA00_013
(Order and components missing -- a Hammer2 issue?  will find out).

https://numeris.canadiana.ca/view/numeris.RD_2004_FA00_013
http://iris.tor.c7a.ca:5984/_utils/index.html#database/copresentation/numeris.RD_2004_FA00_013

jjfriedman:
To clarify, is this only *new* born-digital PDFs or existing ones as well?

RussellMcOrmond:
Both.

We have some existing born digital PDFs in the Numeric project that won't work on the new access platform unless/until we add that functionality.  An example is provided:

* New Access demo: https://numeris-demo.canadiana.ca/view/numeris.RD_2004_FA00_013
* Legacy access we are trying to decommission: https://numeris.canadiana.ca/view/numeris.RD_2004_FA00_013


We also don't have adding new born digital PDFs (or anything other than images) in the roadmap for the new/transitional access platform, so if we temporarily add support for those legacy PDFs they will be the only ones that work until we add that new functionality to Archivematica and related tools (if such a feature is needed).  The ability to ingest those born digital PDFs was temporarily hacked into the preservation platform to fulfil the requirements of that project, and that ugly hack was since removed.


We want to keep new and/or unnecessary features to a minimum in order to allow our technology transformation to happen smoothly, and I'm trying to determine if born digital PDFs is a feature that we need to support in the platform during this transition.

RussellMcOrmond:
Note to @DivyaKrishnan26 and @brittnylapierre ,

This work isn't yet scheduled, but I thought I'd make you aware of this issue.

I put "numeris.RD_2004_FA00_013" into the lookup at https://access.canadiana.ca/ and got to https://access.canadiana.ca/object/69429/m0028pc2tf1g

Changing slug and label, unpublishing and deleting are the only things that can be done.  There is an error in the console: Cannot read properties of undefined (reading 'length')

The "Manage Content" tab should probably not be enabled, as adding a canvas would likely cause everything to break (only from="canvases" can have the `canvases[]` array, while from="pdf" has a `pageLabels[]` array).

[...]

PR #392, opened 2021-12-02
Data cleanup

### LINKED ISSUE #379: Fall cleaning: `data`
- [ ] Remove aliases from the spec, for now. We can re-add them when we need to, or come up with an alternative (like an `extraSlugs` array in manifest/collections).
- [ ] Decouple canvases from the access object spec. They are different enough that a lot of code we write has to work around them, for no benefit.
- [ ] Clean up exports; Vite might be able to handle wildcard exports now.
- [ ] Create a new `pdf` type to assist with #384

SaschaAdler:
Closes #379.

- [x] Remove aliases from the spec, for now. We can re-add them when we need to, or come up with an alternative (like an extraSlugs array in manifest/collections).
- [x] Decouple canvases from the access object spec. They are different enough that a lot of code we write has to work around them, for no benefit.
- [x] Create a new pdf type to assist with #384
- [x] `type` view
- [x] making sure nothing breaks in the admin tools

I'll sort out exports, maybe, when we update SvelteKit/Vite.


RussellMcOrmond:
Related to the "pdf" type, I am wondering if you can create a view that would index and count access documents by type?

Output would be similar to: https://admin.canadiana.ca/couchview/access/tree/pdf?group=1&range=false&reduce=true

* skip documents with an _design/ at the beginning of the ID, and those aren't access documents.
* Output "none" if the type field is missing  (we want to notice these, as they aren't valid and shouldn't exist)
* There would then be 3 types at the moment, possibly more in the future.

SaschaAdler:
A couple things:

- I haven't deployed the new view to production yet. Seems to work fine locally.
- I had to change a bunch of types around (primarily) across the repo. Hopefully things are a little cleaner. Editors seem to load alright, but I'm not sure if any other functionality isn't working.
- I removed the "unassign slug" test because it breaks the new data model and therefore fails. I haven't implemented actual deletion of access objects yet. That can probably wait for the next PR.

SaschaAdler:
I'll revert the slug changes. Everything else shouldn't affect anything except for manifests without `canvases` fields.

RussellMcOrmond:
This mango query returns what we would expect, which is the 738 born digital PDF's we will be changing the type of.

```
{
   "selector": {
      "type": "manifest",
      "canvases": {
         "$exists": false
      }
   }
}
```

RussellMcOrmond (approved):

RussellMcOrmond:
BTW:  This returns no results, also as expected.

```
{
   "selector": {
      "type": "manifest",
      "from": "pdf",
      "canvases": {
         "$exists": true
      }
   }
}
```

Issue #403, opened 2021-12-15
Assumptions made about transparent IDs in existing metadata bus


RussellMcOrmond:
This is not part of PASplit, but something we want to think about soon.

There is code in the current metadatabus which makes assumptions about the transparent ID's.

For instance, when Press deletes the older page documents it does so by assuming that the ID's are in the format of  $manifestSlug, a period, and then something else.  If ever there was a slug that had two periods, and the first part matched a different slug, then this pattern would delete all the records for the slug with two periods.

https://github.com/crkn-rcdr/CIHM-Meta/blob/d32e0454bd50fa3ed7fff5b20e0d9d57aac3d947/lib/CIHM/Meta/Press/Process.pm#L375-L387

* Short term
  * We ensure staff know to continue to use the same method for allocating transparent IDs as they did in the past, meaning *exactly one period* in any "multi-part" collection (AKA: series) or manifest.  "unordered" collections aren't pushed into the old metadatabus, so their TIDs won't cause any issues.
  * I don't know if it is worth baking this limitation into the assigning of slugs, or dealing with this in documentation
* Longer term: replace the metadata bus.  This is a medium sized project, depending on which direction we take.
  * I believe we have two broad directions:
    * Upgrade the existing metadatabus to use OIDs rather than TIDs to manage records, avoiding assumptions about how the IDs are formed.
    * Replace the metadatabus (Hammer2, press, solrstream) with new tools that interacts directly with Solr and Presentation documents, using the OIDs to manage records rather than the TIDs (even if TIDs are the primary keys in some databases).  This may require upgrading CAP accordingly.  This is also a good time to split the Solr search cores between document-level search and page-level search.  #340

SaschaAdler:
I think splitting Solr search cores and only using OIDs to reference canvases/pages is a useful first step, regardless of which longer-term direction we choose to go in. CAP can be easily modified to support this, as there are currently no searches that take place simultaneously across series/issues/monographs and pages.

RussellMcOrmond:
I need to evaluate the current state of this, as quite a bit of work has been done on using NOIDs in cosearch/copresentation as part of https://github.com/crkn-rcdr/cihm-metadatabus/issues/23

RussellMcOrmond:
The current metadatabus uses noids when removing old documents from cosearch/copresentation, so this particular issue has been dealt with.

Software which replaces the metadatabus in the future will need to take different identifiers into consideration.

Issue #400, opened 2021-12-09
Hammer2 doesn't clear out old record correctly.


RussellMcOrmond:
When a slug is changed, Hammer2 is supposed to look up the NOID to ensure the record at the old transparent ID is removed before the new record as the new transparent ID is created.

This is apparently not happening.  @BeS4  changed the ID for oocihm.lac_reel_t16366 , but https://heritage-demo.canadiana.ca/view/oocihm.lac_reel_t16366/1?r=0&s=1 still exists.

RussellMcOrmond:
http://iris.tor.c7a.ca:5984/_utils/index.html#database/copresentation2/oocihm.lac_reel_t16366 has:

"noid": "69429/m0zc7rn3130s"
 
That ID currently has the slug of "BESoocihm.lac_reel_t16366"
https://access.canadiana.ca/object/69429/m0zc7rn3130s

https://heritage-demo.canadiana.ca/view/BESoocihm.lac_reel_t16366/1?r=0&s=1 also exists.


FYI to @brittnylapierre

RussellMcOrmond:
Apparently this is a TODO in the code that was never worked on, and which also needs to be changed to support slugs being mandatory and the slug changing rather than being removed.


https://github.com/crkn-rcdr/CIHM-Meta/blob/d32e0454bd50fa3ed7fff5b20e0d9d57aac3d947/lib/CIHM/Meta/Hammer2/Process.pm#L162-L175

brittnylapierre:
I fixed the example with the issue that Russell provided for Beth, so you won't see the problem anymore with the manifests above. We'll have to replicate the issue with our own manifest to test a fix.

RussellMcOrmond:
We have a problem in our logic currently, which we will want to decide which way we wish to fix.

The Hammer2 code assumed that the only slugs that would change are those which were already unpublished.  It is unpublishing an existing slug that actually removes it from the cosearch and copresentation databases.

If you change the slug for something published, then the system needs to look up the noid and ensure that the old slug is unpublished as part of publishing the new one.

We either want to enhance the logic in Hammer2, or enhance the logic in the web interface to disallow changing the slug without first unpublishing the old slug - and ensuring that Hammer2 -- AKA: what uses updateInternalmeta{} is finished..

RussellMcOrmond:
[...]

PR #404, opened 2021-12-15
Cleanup of old documents (slug changing, etc).

RussellMcOrmond:
Two related issues seem to be pointing at a related solution:

* #400 where changing the slug leaves the document at the old slug intact
* #403 where the example of Press making assumptions about IDs can be fixed by actually using the parent ID of pages.

The relevant OID and TID are both in the cosearch and copresentation documents, and we can make use of this fact in any cleanup.


RussellMcOrmond:
The design documents for these views remain under `deprecated/Databases`.  This is forward-thinking in that there will be a future project to replace the current metadatabus, but at the moment these databases are still active.

SaschaAdler (approved):

PR #406, opened 2021-12-15
Ensure the process order of a DMD task always matches the table order

### LINKED ISSUE #405: DMD Task process order depends on id value (because of map usage.)
Use the items array instead of the map to determine process order.

brittnylapierre:
Fixed so that the dmd task item list is used for iteration. 

The table was being processed in reverse for Natalie, which was being caused by the ids being in descending order!!

Closes #405 

We can see if we can simplify the dmd task logic in the future


RussellMcOrmond (approved):
As long as the loop is complete, I don't have any comments about the order.

Issue #231, opened 2021-08-10
Reinstate storage policy doc


SaschaAdler:
I threw it out by accident along with accessor. It should live in `/docs`. The text is as follows:

# Access Object Storage scheme

The preservation-access platform split means that the generation and maniuplation of images and metadata on the access platform can no longer be thought of as a by-product of the preservation process. As such, a scheme for storing access images and metadata manually derived from what is available in the preservation platform is needed.

CIHM::Smelter will accept access deposit requests from a queue, retrieve the relevant data from the preservation platform, split and manipulate preservation METS records, and store the results in Swift using the following scheme.

* /v1/AUTH_crkn/access-metadata/$noid
  * /dmdMARC.xml
  * /dmdDC.xml
  * /dmdISSUEINFO.xml
  * /ocrALTO.xml
  * /ocrTXTMAP.xml
* /v1/AUTH_crkn/access-files/$noid.$ext

Couch databases for collections, manifests, and canvases will be populated with the required information to link these objects together.

$noid is [a noid](https://metacpan.org/pod/distribution/Noid/noid), generated by our [Noid service](https://github.com/crkn-rcdr/noid). It can identify a collection, manifest, or canvas.

Note: Swift containers are generally created under AUTH_crkn, and access controls used to determine what specific users have access.  Different applications should be using different user IDs so that we can grant only the access required for that specific application.

PR #424, opened 2021-12-17
Critical updates based on user feedback that make workflows clearer - PDFs & DMD fixes

brittnylapierre:
Please see the list of issues included in this PR. They're mostly superficial changes to how we display information in the UI so that it's easier to understand based on the goals people have when they come to the app.

The only thing that required me to touch lapin and the database handler was the delete method on the back-end. I had to get it set up because it's an important step someone has to take when they are re-importing a package into the access platform.

Edit - I ended up having to wait for things and added the PDF support in the editor and a DMD bug fix in this branch too, which required backend changes. I'll remember to make new branches next time

I will continue with the rest of the changes in a couple new PRs. (One for editor, one for the remaining DMD changes) 

I would like to show Beth the import changes after making a new image to get some feedback on them.


SaschaAdler:
re: deleting objects, and correct me if I'm wrong, @RussellMcOrmond 

Deleting an access object should also delete associated files in Swift; specifically, any associated descriptive metadata files and PDFs. Objects with these associated files will have fields in the database that indicate this.

RussellMcOrmond:
> Deleting an access object should also delete associated files in Swift; specifically, any associated descriptive metadata files and PDFs. Objects with these associated files will have fields in the database that indicate this.

Correct.

Anything associated with that OID in `access-files` and `access-metadata` should also be deleted.
See: #231 

If some were forgotten that is OK for now as I will write a walker to garbage collect things that we didn't get right in the past, but best to clean up right away.

brittnylapierre:
> > Deleting an access object should also delete associated files in Swift; specifically, any associated descriptive metadata files and PDFs. Objects with these associated files will have fields in the database that indicate this.
> 
> Correct.
> 
> Anything associated with that OID in `access-files` and `access-metadata` should also be deleted. See: #231
> 
> If some were forgotten that is OK for now as I will write a walker to garbage collect things that we didn't get right in the past, but best to clean up right away.

Ah thanks @SaschaAdler and @RussellMcOrmond ~ I can add that in tomorrow then request another review!

brittnylapierre:
@SaschaAdler @RussellMcOrmond Sorry I got a bit stuck.

Would this be the place that tells me if there's a pdf to delete?

(Was looking at the access database just to see the fields that tell me what to delete - this is an example, not something I'm actually trying to delete:)
 "ocrPdf": {
    "size": 128232,
    "path": "ooe.sas_19951212AEP/data/sip/data/files/ooe.19951212AEP.pdf"
  }

I am not sure if the ocr pdf tells me that there is a pdf to delete. If so, how do I delete it? 

I am not sure how to translate the path into what's needed for swift (again the above isn't my testing file but just an example.) 

I tried to look up the manifest's noid with the 'pdf' extension in access-files too and it's saying it does not exist. So that makes me think either the ocrPdf isn't the right field to tell me if there is a pdf to delete, or that I am searching in the wrong place.

I know the metadata part pretty well from my work on the dmd tasks. I think I need a bit more direction on the access-files side of things though.

Thanks!

RussellMcOrmond:
* In a `fileRef` and an `imageRef` there is a transition happening.
  * If there is a "path" field, then you just ignore.  That is a pointer into the preservation system, which shouldn't be touched.
  * If there is an `extension` field then there will be a `/v1/AUTH_crkn/access-files/${noid}.${extension}` that you need to delete (use the configuration for the account and container -- it is the $noid.$extension that is the Swift object ID)
* the `file` and `ocrPdf` fields are a fileRef, and one or both of these may exist depending on which type of access document you are trying to delete. 
* All the access objects are of `AccessObjectTrait` which, if it has a `dmdType` field, then has a `/v1/AUTH_crkn/access-metadata/${noid}/dmd${DMDTYPE}xml ` you need to delete ($DMDTYPE is the upper-casing of the `dmdType` field so `dmdType="marc"` means the Swift Object is `${noid}/dmdMARC.xml`. )


The other files we store in access-* containers in Swift are Canvas related based on Canvas OIDs, so you don't need to worry about them at the moment.  (OCR XML in access-metadata, OCR PDF in access-files, and images in access-files).

That transition thing was needed as it takes time to copy all the files, and I wanted us to be able to start testing software before the files were copied.  It is the intention that the `path` field in `fileRef` and `imageRef` will eventually go away. 

Hope that helps.

brittnylapierre:
Thanks @RussellMcOrmond ! I'll make those changes.

I ended up having to wait a bit for the deletion work and I should have branched off for these but I ended up doing the pdf changes here too. So I can open up pdfs in the editor now.
https://access-dev.canadiana.ca/object/edit/69429/m0000000474p 

I also fixed the dmd task issue where the dmdType wasn't being updated. I tested with two of the items that Natalie pointed out, here's one and you can see the subjects match now:
https://www-demo.canadiana.ca/view/oocihm.65600/1?r=0&s=1
https://www.canadiana.ca/view/oocihm.65600/1?r=0&s=1

I'll remember to make new branches while I wait next time~

RussellMcOrmond (approved):
I only skimmed, and nothing jumped out at me.

Once we get past PASplit, we'll hopefully have some time to look back and do some reviews of the tools overall.   I know I have some suggestions from a systems/DB point of view.

brittnylapierre:
Yeah I'll be excited to be able to improve things wherever possible! We have the cleanup issues too that will remind us to go through things. I will merge this to and build a new image.

PR #434, opened 2021-12-23
Importer bug-fix

### LINKED ISSUE #430: Slug resolver may be searching for aip id on page load even if slug is different (on smelter view pages)
Screenshot from 2021-12-23 16-10-23

Pointing me to the manifest with the aip id as a slug not something called bl.test3

to duplicate try going to packages with recent activity and running an import on something

brittnylapierre:
Closes #430 

To test please go to packages with recent activity -> find a manifest with an unavailable slug -> enter a slug that is available -> select it -> press import -> go to queue and see that the slug you entered is marked as available. 

I can make this queue view have a non-editable slug after fix is confirmed too, since it really shouldn't be an input there.


RussellMcOrmond (approved):
Logic looks good.  I didn't build and run, but trust that it does.

PR #437, opened 2022-01-03
Adding update functions required for Importing OCR from legacy preservation.

RussellMcOrmond:
https://github.com/crkn-rcdr/Access-Platform/issues/368

Also a good time to set up replacement update functions for Smelter as well.


DivyaKrishnan26 (approved):

SaschaAdler (approved):
There are some things in here that could stand to be combined (i.e. replace `updateObject` and `updateGenericObject` in the prelude with one thing that does the right thing to some `ProcessUpdate` in some document) but that's to sort out later.

PR #442, opened 2022-01-06
Adding new error messages to the metadata updater

brittnylapierre:
We were able to determine a list of error strings that would be helpful for the users. I'll be going through the app for the DMD task stuff and adding in or editing those errors into the admin app code and into the dmd task specific code in lapin (which will be moved out eventually)


brittnylapierre:
Hey @SaschaAdler is there a better way to catch errors in the input step and have a specific error message like how I have there now? I was wondering if there was a better way for the code to know the input type too?

What I have there does what I want but I'm not sure if it's the cleanest way to do it. I'll be editing a lot of the requests to do something similar so I want to ask before I change a lot of stuff.

PR #445, opened 2022-01-10
Rename the 'Import History' table to 'Last Import Status'

brittnylapierre:
I ended up adding in the re-import or copy option for the search results. Inititally after talking with the heritage team they said they probably wouldn't use the additional import/new copy function too much, but then we were all talking today and it came up that people would in fact be making additional imports of manifests as temporary manifests to update OCR.

So, I got on setting this up quickly.

When the slug is not available, they have clear (I hope, we'll see) options for either replacing an existing import (re-import) or making an additional manifest from the package (copy)

Here's what it looks like:
Screenshot from 2022-01-10 12-01-43
Screenshot from 2022-01-10 12-01-51
Screenshot from 2022-01-10 12-02-05

### DIFF
--- a/services/admin/src/lib/components/dipstaging/DipstagingLookupTable.svelte
+++ b/services/admin/src/lib/components/dipstaging/DipstagingLookupTable.svelte
@@ -24,6 +24,7 @@ This component shows the results of a dipstaging find-package(s) request or a vi
   import NotificationBar from "../shared/NotificationBar.svelte";
   import ExpansionTile from "../shared/ExpansionTile.svelte";
   import { afterUpdate } from "svelte";
+  import ToggleButtons from "../shared/ToggleButtons.svelte";
 
   /**
    * @type {ImportStatus[]}
@@ -68,6 +69,8 @@ This component shows the results of a dipstaging find-package(s) request or a vi
    * @type {any} A map from ImportStatus.id => if the slug of the item's slug is available.
    */
   let slugMap: any = {};
+
+  let modeMap: any = {};
 
   let error = "";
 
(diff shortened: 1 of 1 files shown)


RussellMcOrmond:
What we discussed today isn't a re import or a copy.  It is a separate ID in preservation that should go to a separate temporary ID in access.

Then the "Edit in Access" tool would be used to add to the correct ID the canvases from the temporary ID, and then the temporary ID manifest can be deleted.

RussellMcOrmond:
I suggest "Import history" be called "Last Import Status" or similar.  The word history suggests that there could be more than one previous status.

There is also a [cancelSmelt](https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/dipstaging/design/access/updates/cancelSmelt.js) which can be used to clear that last status, so it doesn't continue to show up in https://access.canadiana.ca/smelter/status/1/10/all if there is no longer an issue.  That page should be clear when the operator doesn't need to know the message any longer.

brittnylapierre:
Sounds good @RussellMcOrmond thanks for explaining. 

I'll comment out the copy stuff for now and I'll change the label to 'Last Import Status.'

Did you want me to add a button to the queue page to allow people to cancel a smelt? Or just a button to the Last Import Status to clear it? Or neither? Lol thanks!

RussellMcOrmond:
> Did you want me to add a button to the queue page to allow people to cancel a smelt? Or just a button to the Last Import Status to clear it? Or neither? Lol thanks!

Or both?  It is on https://access.canadiana.ca/smelter/status/1/10/all where the errors are most visible, so a "cancel" or "clear" there would likely be ideal.   What that update does is clear out the doc.smelt field, so it both removes from the queue if it was in, but it also clears any error.  You'll want to word it in the interface to make this clear to operators.

Note: It doesn't stop processing already happening, only removes from the queue or removes past status.

brittnylapierre:
Ahh right that makes most sense. So, I'll add it to https://access.canadiana.ca/smelter/status/1/10/all as well as the Last Import Status table!

I start off with 'Clear Last Import Status'  as the label, I think!

brittnylapierre:
Actually @RussellMcOrmond I made the button a new issue and I'll just ask to merge the table name change here from 'Import History' to 'Last Import Status.' Feel free to edit the priority of this change and location in the backlog on the Developers Work board.

PR #450, opened 2022-01-13
Simplify the deletion checks

brittnylapierre:
Now no matter what type an object is:

If it has the file.extension, it will delete the file

If it has the ocrPdf.extension, it will delete the ocr

And if it has the dmdType property, it will delete the metadata file.

Let me know if you spot anything else!

### DIFF
--- a/packages/lapin-router/src/routes/accessObject.ts
+++ b/packages/lapin-router/src/routes/accessObject.ts
@@ -130,7 +130,7 @@ export const accessObjectRouter = createRouter()
         }
 
         // Check if the file field exists, if so we know to delete it
-        if ("file" in accessObj && accessObj?.file?.extension) {
+        if ("file" in accessObj && accessObj.file?.extension) {
           const fileName = `${accessObj.id}/${accessObj.file.extension}`;
           try {
             await ctx.swift.accessFiles.deleteObject(fileName);
--- a/packages/lapin-router/src/routes/accessObject.ts
+++ b/packages/lapin-router/src/routes/accessObject.ts
@@ -8,7 +8,6 @@ import {
   toPagedManifest,
   toPagedCollection,
   AccessObject,
-  Manifest,
 } from "@crkn-rcdr/access-data";
 import { createRouter, httpErrorToTRPC } from "../router.js";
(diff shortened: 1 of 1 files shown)


packages/lapin-router/src/routes/accessObject.ts
```diff
@@ -118,32 +117,29 @@ export const accessObjectRouter = createRouter()
...
-            const fileName = `${manifest.id}/${manifest.ocrPdf.extension}`;
-            try {
-              await ctx.swift.accessFiles.deleteObject(fileName);
-            } catch (e: any) {
-              console.log(e?.message);
-            }
+        /* Delete files from access-files */
+
+        // Check if the OCR field exists, if so we know to delete it
+        if ("ocrPdf" in accessObj && accessObj.ocrPdf?.extension) {
```
RussellMcOrmond, line 123: ES question:  If you are using `ocrPdf?` doesn't that mean that there won't be a failure if `ocrPdf` doesn't exist, and thus you don't need the first check?  Doesn't change the logic, so just me being curious.

RussellMcOrmond (approved):
I looked up https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

There is some code simplification you can do if you wish.

Logic looks great.


In the future (not now, not a priority) we might want to think about error handling, and if we want to report to users.

More complex logic to consider in the future:

* Try to delete
* if it succeeds then update the access document in memory accordingly.
* If it fails, don't continue : save access document if it changed, and report error to operator.
* Only delete the access document if all Swift object deleting was successful.

packages/lapin-router/src/routes/accessObject.ts
brittnylapierre: Ah thank you Russel you are right there, really I was just checking that extension exists so I don't need the ?

The only reason why I'm using 'in' is because using the ocrPdf?.extension does not clear a warning about ocrPdf not existing on all kinds of accessObjects.

brittnylapierre: Ah it seems when I remove the ? I get a warning that it is possibly undefined, nevermind!

PR #457, opened 2022-01-17
Adding OCR Batch schemas

brittnylapierre:
Followed what was set up for the other databases. I did end up having to create the ocr database through the dashboard before kivik populated it with my fixtures. I tried to make tests based off of the dmd task tests that were already set up.


RussellMcOrmond:
In this PR, can you create "request", "cancel" and "update" documents for each of the fields?
https://github.com/crkn-rcdr/Access-Platform/tree/main/services/couchdb/dipstaging/design/access/updates
Notice:
https://github.com/crkn-rcdr/Access-Platform/blob/24be76e83434ed6cee4a5de8539b4765[…]ervices/couchdb/dipstaging/design/access/updates/updateSmelt.js

This could be enhanced to not only check it is a string, but that it isn't a blank string.  We could decide that a blank string isn't something we ever want to store.

RussellMcOrmond:
Ignore what I suggested about the `message` field being optional, and thus blank for a silent success.

I refreshed my memory of https://github.com/crkn-rcdr/Access-Platform/blob/main/packages/data/src/util/ProcessUpdate.ts

Allowing the `message` field to be optional isn't part of a `SucceededProcessResult` or `FailedProcessResult`.

We can leave the spec that way and always have message="" for a silent success.  This would be handled in the _update documents (instead of the delete of the field, have it set field to "").

Whatever is decided, we can do some "spring cleaning" later to have `SmeltProcess` be the same, so it doesn't have to be separate.  Currently it has a unique spec because current code has more fields being optional.

brittnylapierre:
Ah oops @RussellMcOrmond I ended up making the SucceededProcessResult or FailedProcessResult have optional messages :o And added the tests for those cases. Is that okay too? I can revert that change if you prefer.

brittnylapierre:
I ended up making a custom process result for the batches that can have no message, we can update the general process result and get rid of the smelt and ocr custom ones some future time

brittnylapierre:
I made tests in /Access-Platform/packages/couch-utils/src/handlers/ocr.spec.ts to test all the update design docs and they're passing

packages/data/src/ocr/Batch.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+const { isValid: isValidImportSucceeded } = tester(ImportSucceededOcrBatch);
+
+const USER: User = { name: "User McGee", email: "mcgee@crkn.ca" };
+
+const timestamp = (d: number) => {
+  return new Date(d).toISOString().replace(/.\d+Z$/g, "Z");
+};
+
+const now = timestamp(Date.now());
+const then = timestamp(Date.now() - 100);
```
RussellMcOrmond, line 25: The update documents will be using ISO dates, so probably a good double-check to use the same date format for the testing.

packages/data/src/ocr/Batch.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  goodExportFailed
+);
+
+const goodExportFailedNoMessage: ExportFailedOcrBatch = {
+  ...goodExportWaiting,
+  id: "processing-failed",
+  exportProcess: {
+    requestDate: then,
+    succeeded: false,
+    processDate: now,
```
RussellMcOrmond, line 72: No or blank message on a `succeeded: false` is actually a bug in the microservice, and should never happen.

RussellMcOrmond (changes requested):

RussellMcOrmond:
I saved before finishing.  I added a Queue and status.  While I tested by modifying some of the test documents in the database, you may want to add some additional fixtures and tests for them.  Adjust them as you feel appropriate.

packages/data/src/ocr/Batch.spec.ts
brittnylapierre: Okay! I've changes them to iso strings

packages/data/src/ocr/Batch.spec.ts
brittnylapierre: Okay! I made it so that the failed versions use the regular failed process update, so messages are required

RussellMcOrmond (approved):

PR #458, opened 2022-01-18
Indexing `manifest_noid` rather than `pkey`.  

RussellMcOrmond:
This will be used to remve pages when a manifest is removed, rather than making assumptions about the structure of a slug (there may be more than one `.`)

See https://github.com/crkn-rcdr/Access-Platform/issues/400

### DIFF
--- a/services/couchdb/copresentation2/design/access/views/manifest_noid.js
+++ b/services/couchdb/copresentation2/design/access/views/manifest_noid.js
@@ -1,8 +1,8 @@
 // Used by Press2
 module.exports = {
   map: function (doc) {
-    if ("pkey" in doc) {
-      emit(doc.pkey, null);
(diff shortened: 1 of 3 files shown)


RussellMcOrmond:
Idea is simple: one view finds the manifest and the other finds the pages within that manifest.  I could combine them into one view which would return both the manifest and any of its pages.

Press currently makes an assumption about the ID's which may not be true, and it is possible (even if statistically unlikely) that the current pattern could delete other manifests rather than only components.  The current lookup assumes that a slug could only ever have a single period, and that any second period must by definition be a page.

https://github.com/crkn-rcdr/CIHM-Meta/blob/93570b08fdea5491fec2be9b6e477d451a17d9f3/lib/CIHM/Meta/Press/Process.pm#L375-L387


Thoughts?

PR #461, opened 2022-01-18
Fix the issue of non propagating changes

brittnylapierre:
Hmm the force update is getting called when membership changes, then the same method (updateObject) is already getting called in the editObject design doc, so theoretically every update should be triggering the microservice.

The updateObject method does this:

```

/**
 * Updates the fields that change when an Access Object is edited. Creates an internalmeta request.
 * @param {Record<string, string>} doc The Access Object's document.
 * @param {{name: string; email: string} | undefined} user The user who triggered this update, if one exists.
 */
exports.updateObject = (doc, user) => {
  const now = exports.timestamp();
  if (
    typeof user === "object" &&
    typeof user.email === "string" &&
    typeof user.name === "string"
  ) {
    doc.staff = { by: user, date: now };
  }
  doc.updated = now;
};
```

Does that look correct? It is what forceUpdate and editObject both call. 

It doesn't try to set the updatInternalmeta object at all.

When I update it to:

```
/**
 * Updates the fields that change when an Access Object is edited. Creates an internalmeta request.
 * @param {Record<string, string>} doc The Access Object's document.
 * @param {{name: string; email: string} | undefined} user The user who triggered this update, if one exists.
 */
exports.updateObject = (doc, user) => {
  const now = exports.timestamp();
  if (
    typeof user === "object" &&
    typeof user.email === "string" &&
    typeof user.name === "string"
  ) {
    doc.staff = { by: user, date: now };
  }

  if (doc.updateInternalmeta) delete doc.updateInternalmeta;
  doc.updateInternalmeta = { requestDate: now };

  doc.updated = now;
};
```

The changes are reflected on the demo site as expected in aprox 15 mins.

Should this be considered the fix? 

Do the copresentation and cosearch databases get updated periodically even if the updateInternalmeta isn't updated? I ask because if not, if the collection label changes, every member would need to be forced to update to show the new label of the collection it is in. 

If it does update periodically even if the updateInternalmeta isn't changed, we don't have anything to think about, since eventually the label will be fixed. If it does we might need to think on this, because some collections have many many members and calling forceupdate on all of them at once is probably not good?

Thanks!

### DIFF
--- a/packages/lapin-router/src/routes/collection.ts
+++ b/packages/lapin-router/src/routes/collection.ts
@@ -143,9 +143,7 @@ export const collectionRouter = createRouter()
     input: EditInput.parse,
     async resolve({ input, ctx }) {
       try {
-        const result = await ctx.couch.access.editCollection(input);
-
-        return result;
+        return await ctx.couch.access.editCollection(input);
       } catch (e) {
         throw httpErrorToTRPC(e);
       }
--- a/packages/lapin-router/src/routes/collection.ts
+++ b/packages/lapin-router/src/routes/collection.ts
@@ -143,7 +143,9 @@ export const collectionRouter = createRouter()
     input: EditInput.parse,
     async resolve({ input, ctx }) {
       try {
-        return await ctx.couch.access.editCollection(input);
+        const result = await ctx.couch.access.editCollection(input);
+
+        return result;
       } catch (e) {
         throw httpErrorToTRPC(e);
       }
--- a/packages/lapin-router/src/routes/manifest.ts
+++ b/packages/lapin-router/src/routes/manifest.ts
@@ -200,11 +200,7 @@ export const manifestRouter = createRouter()
     input: EditInput.parse,
     async resolve({ input, ctx }) {
       try {
-        const result = await ctx.couch.access.editManifest(input);
-
-        await ctx.couch.access.forceUpdate(result.id);
-
-        return result;
+        return await ctx.couch.access.editManifest(input);
       } catch (e) {
         throw httpErrorToTRPC(e);
       }
--- a/packages/lapin-router/src/routes/manifest.ts
+++ b/packages/lapin-router/src/routes/manifest.ts
@@ -200,7 +200,11 @@ export const manifestRouter = createRouter()
     input: EditInput.parse,
     async resolve({ input, ctx }) {
       try {
-        return await ctx.couch.access.editManifest(input);
+        const result = await ctx.couch.access.editManifest(input);
+
+        await ctx.couch.access.forceUpdate(result.id);
+
+        return result;
       } catch (e) {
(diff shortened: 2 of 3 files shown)


RussellMcOrmond:
`updateInternalmeta` is what causes Hammer2 and (for now until merged in) Press to do their work and update the `cosearch` and `copresentation` caches.

We are temporarily assuming that only `series`, `issues` and `monographs` can exist -- specifically that any given manifest can only ever be a member of 0 or 1 multi-part collections.  We generate the concept of a "parent" for issues, and then get data from that parent.

In order for an issue to notice that  the label of its parent has changed, it needs to have `updateInternalmeta`  be forced.  It is all background processes that staff don't have to wait for.


See what copies the `plabel` in https://github.com/crkn-rcdr/CIHM-Meta/blob/93570b08fdea5491fec2be9b6e477d451a17d9f3/lib/CIHM/Meta/Press/Process.pm#L689-L728

Same for when a member of a multi-part collection is published or unpublished : the collections it is a member of needs to be processed as well.

The admin tools and the metadatabus doing work ahead of time is all about reducing the amount of work that needs to be done when a patron requests something.  We don't want to optimise the admin tools at the expense of fast patron access, and all the admin tools need to do is the forceUpdate() so that the metadatabus knows to update the caches.. That is so when CAP is asked to look at an issue that the `copresentation` document has everything it needs in a minimal number of database requests, rather than it doing the lookups at that time.

The admin tools should only be indicating to microservices that work needs to be done, and to report to the users when it is done.  The actual work shouldn't  be done by the admin tool directly (it isn't a web thing, and shouldn't cause the browser to delay).

`cosearch` and `copresentation` are caches to speed up access to data when patrons ask for it.  One is fed into Solr, and the other is read directly by CAP.


For the longer term discussion there are two documents in the[ Access-Platform1](https://crkn.sharepoint.com/:f:/r/sites/platform/Shared%20Documents/Projects/Access%20Platform1?csf=1&web=1&e=6f4iL1) documention folder it is worth reading that discuss related problems:

* Issues with Issueinfo (I wrote)
* Issue display labels on the new access platform  (Sascha wrote)

Longer term we will want to cache sufficient data to efficiently display, but remove the concept of a "parent" as manifests will be able to be members of multiple multi-part collections.  If a breadcrumb needs to be displayed, it should be able to be displayed from data within the relevant `copresentation` document.  Enhancing the schema  to handle this is outside the scope of the current project.

RussellMcOrmond (approved):
I'm approving the small change.

Additional calls to this update will be needed to deal with

* collections an edited manifest is part of
* members of a multi-part collection that was edited (including label)
* members of an unordered collection that was edited (label doesn't matter, but public/private and slug does).

brittnylapierre:
Okay! @RussellMcOrmond I'll create a new issue for tracking those three improvements

PR #466, opened 2022-01-19
Force updates for collection members when important fields change, and force update parent collections when manifests updated

brittnylapierre:
I tested yesterday with the news collection, renaming it to newspapers. As well with my own small collection. Everything forked as expected. I tested the generalisation with my small collection. I'm going to run it again to reset the news collection too.

### DIFF
--- a/packages/couch-utils/src/DatabaseHandler.ts
+++ b/packages/couch-utils/src/DatabaseHandler.ts
@@ -168,22 +168,43 @@ export class DatabaseHandler<T extends Document> {
   }
 
   async forceUpdateMany(
-    docs: {
-      _id: Noid;
-      updateInternalmeta: {
-        requestDate: Timestamp;
-      };
-    }[]
+    ids: any[] // todo: make string once object list item id is not optional
   ): Promise<DocumentBulkResponse[]> {
     /**
      * https://docs.couchdb.org/en/stable/api/database/bulk-api.html#db-bulk-docs
      * For updating existing documents, you must provide the document ID, revision information (_rev), and new document values.
      */
 
-    console.log("bulk: ", docs);
+    const date = new Date().toISOString().replace(/.\d+Z$/g, "Z");
+
+    const bulkUpdateDocs: {
+      _rev: string;
+      _id: Noid;
+      updateInternalmeta: {
+        requestDate: Timestamp;
+      };
+    }[] = [];
+
+    const fetchRes = await this.db.list({ keys: ids });
+
+    fetchRes.rows.map((row) => {
+      console.log("row", row);
+      //TODO: how to get this to only replace the feilds in here not the full doc?
+      if (row.id && row.value?.rev) {
+        bulkUpdateDocs.push({
+          _id: row.id,
+          _rev: row.value.rev,
+          updateInternalmeta: {
+            requestDate: date,
+          },
+        });
+      }
+    });
+
+    console.log("bulk", bulkUpdateDocs);
 
     return await this.db.bulk({
-      docs,
+      docs: bulkUpdateDocs,
     });
   }
 
--- a/packages/couch-utils/src/DatabaseHandler.ts
+++ b/packages/couch-utils/src/DatabaseHandler.ts
@@ -124,14 +124,6 @@ function chunkArray(array: any[], n: number) {
 }
 
 /**
- * A helper method to slow down the rate of requests going to the backend. Causes the script to pause for 'ms.'
- * @param ms
- */
-function sleep(ms: number) {
-  return new Promise((resolve) => setTimeout(resolve, ms));
-}
-
-/**
  * Handler for interactions with a CouchDB database.
  *
  * Also handles translating `_id` and `_attachments` to non-underscored versions.
@@ -195,6 +187,7 @@ export class DatabaseHandler<T extends Document> {
   /**
    * Updates all the objects in ids
    * @param ids Id strings for the objects to be updated
+   * @param changeMethod A function that returns the following: [<new document>, <message>]
    * @returns the result of the bulk update
    *
    * See: https://docs.couchdb.org/en/stable/api/database/bulk-api.html#db-bulk-docs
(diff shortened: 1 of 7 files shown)


packages/couch-utils/src/DatabaseHandler.ts
```diff
@@ -168,22 +168,43 @@ export class DatabaseHandler<T extends Document> {
...
+      updateInternalmeta: {
+        requestDate: Timestamp;
+      };
+    }[] = [];
+
+    const fetchRes = await this.db.list({ keys: ids });
+
+    fetchRes.rows.map((row) => {
+      console.log("row", row);
+      //TODO: how to get this to only replace the feilds in here not the full doc?
```
RussellMcOrmond: CouchDB doesn't let you do this directly with its API (give an array of IDs to do an _update doc against), which is why you want to create a generic procedure to do that yourself which accepts as a parameter the specific function that makes the specific changes you want.


```
function bulkChange (param1, param2, arrayOfIDs, callBackFunction)
{
   
  ...get all the documents for the IDs
  var storarray=[];
  loop through documents  {
     var newdoc = callBackFunction(olddoc);
     if (newdoc) {
      storearray.push(newdoc);
    }
  }
  Push storearray to CouchDB 
}
```

packages/couch-utils/src/DatabaseHandler.ts
```diff
@@ -165,6 +192,53 @@ export class DatabaseHandler<T extends Document> {
...
+          }
+        });
+
+        // Update the database
+        await this.db.bulk({
+          docs: bulkUpdateDocs,
+        });
+
+        // Avoid flooding the database
+        await sleep(10000);
```
RussellMcOrmond, line 235: Can you create a ticket to investigate this in more detail later?

Is it CouchDB or Lapin that was failing when you were sending many requests.   It was likely the number of HTTP requests hitting a queue limit somewhere, and not the size of the request.

Also, what is the HTTP timeout for the library you are using?  Sometimes requests will take time, and that is fine as lapin stuff should have quite long timeouts.

The sleep just seems odd to me, and isn't something I've ever needed to do in any of the PERL code.

brittnylapierre: Ahh I never tried it without the sleep, it sends one request per hundred members. I can try editing the newspaper collection in one go, without the sleep or chunking of it's members into separate requests, to see if it fails.

I would have to ask @SaschaAdler for your HTTP timeout question, the server uses nano https://github.com/apache/couchdb-nano for all of the couch requests

brittnylapierre: Okay if I don't chunk I get:
/repo/node_modules/.pnpm/nano@9.0.3/node_modules/nano/lib/nano.js:175
const errors = new Error(message)
Error: couch returned 400
    at responseHandler (/repo/node_modules/.pnpm/nano@9.0.3/node_modules/nano/lib/nano.js:175:20)
    at /repo/node_modules/.pnpm/nano@9.0.3/node_modules/nano/lib/nano.js:405:13
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
...
  headers: {
    ... database uri...
    statusCode: 400,
    connection: 'close',
    date: 'Thu, 20 Jan 2022 15:02:26 GMT'
  },
...
 ELIFECYCLE  Command failed with exit code 1.

Does that help at all?

brittnylapierre: Chucking with 100 items works, without a sleep. But if I try 500+ it doesn't work~ So I think I can leave the chunking in, but I can leave sleeping out

brittnylapierre: I'll make a ticket for the chunking though if you think!

RussellMcOrmond: It is the sleep I was wondering about.  Thanks.

The chunking will relate to size of documents and the timeout.  Leave the chunking in, and when you have time you can adjust the timeout and the chunking size.  Doing bulk requests of 5000+ documents is possible with CouchDB, but you have to have quite long timeout.  (I regularly use 36000 seconds which is an hour).

Press is only using 3600  -- see https://github.com/crkn-rcdr/CIHM-Meta/blob/main/lib/CIHM/Meta/Press.pm#L85

The default timeout for the Perl library we use is 60 seconds, and I always set it to something much larger.

packages/couch-utils/src/DatabaseHandler.ts
```diff
@@ -165,6 +192,53 @@ export class DatabaseHandler<T extends Document> {
...
+        // Change the data as described in the change method
+        fetchRes.rows.map((row) => {
+          if (row.doc) {
+            const newDoc: any = changeMethod(row.doc);
+            if (newDoc) bulkUpdateDocs.push(newDoc);
+          }
+        });
+
+        // Update the database
+        await this.db.bulk({
```
RussellMcOrmond, line 230: It is possible for bulkUpdateDocs to be empty, so probably best to check for that.  I don't know what CouchDB does in that case.

RussellMcOrmond (approved):

packages/lapin-router/src/routes/collection.ts
```diff
@@ -143,8 +143,40 @@ export const collectionRouter = createRouter()
...
+          const ids: any[] = res.members
+            .filter((member) => typeof member.id !== "undefined")
+            .map((member) => member.id);
+
+          // Don't hold up the response. This will run in the background without causing issues for end users. They don't need to be alerted about any of this in real time. The updateInternalmeta is displayed in the editor.
+          ctx.couch.access
+            .bulkChange(ids, (olddoc: any) => {
+              const date = new Date().toISOString().replace(/.\d+Z$/g, "Z");
+              return {
+                ...olddoc,
```
RussellMcOrmond, line 166: Variable scoping question:  Aren't variables created in the outer scope available?  Can't you define the date above (line 160?) and use it within the function?

I'm also curious if something like the following would be more readable.

```
const date = new Date().toISOString().replace(/.\d+Z$/g, "Z");

ctx.couch.access.bulkChange(ids, (doc: any) => {
  doc.updateInternalmeta = { requestDate: date };
  return doc;
}

```

I don't know ES well, so maybe what you wrote is more readable to those who live in ES.  I'm learning here as well.

brittnylapierre: Ah, I got errors when I wrote it that way, originally, so I ended up cloning the object. Adding in the oldDoc: any removed the error though, so I'll change it to be more readable. Yes I can add the date to the outside! I think I just quickly copied over what I had in the for loop before

packages/couch-utils/src/DatabaseHandler.ts
brittnylapierre: Okay sounds good, I made this ticket for looking into that further: https://github.com/crkn-rcdr/Access-Platform/issues/468

RussellMcOrmond (approved):

PR #467, opened 2022-01-20
Adding internalmeta status information to the editor (no buttons, just status)

brittnylapierre:
I'll only merge this once the force update branch is merged, since I made this branch from that one.

I added a table with the internal meta information. I thought the requestdate and process date was an overkill. 

It shows the status and if there is a message.
Screenshot from 2022-01-20 09-01-38

The status  labels are...

If it is running but not finished:
Access Status: | Currently updating...

If the process succeeded:
Access Status: | Most recent update succeeded!

If the process failed:
Access Status: | Most recent update failed.

Then if there is a message, it will display below in a notification bar.
Screenshot from 2022-01-20 09-01-22

### DIFF
--- a/packages/couch-utils/src/DatabaseHandler.ts
+++ b/packages/couch-utils/src/DatabaseHandler.ts
@@ -193,61 +193,45 @@ export class DatabaseHandler<T extends Document> {
   }
 
   /**
-   * Gets a document from a database. Does not throw an error if the document isn't found.
-   * Returns a the cache status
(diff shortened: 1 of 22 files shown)


RussellMcOrmond:
Only suggestion for access status is to consider displaying the date.  When an update was requested, and when it failed or succeeded, may be quite useful information.

RussellMcOrmond (approved):
Nice and simple.

Issue #464, opened 2022-01-19
Too many AIP ids entered into our searches causes the listfromkeys to return an error


brittnylapierre:
- [x] Chunk large list type queries in the backend
- [x] Do not allow duplicate members to be added to a collection
- [x] Add better checks for weird characters (copy pasting from docs needs to be handled more robustly)
- [x] Add a unique-ify process for the slugs in the search box

PR #470, opened 2022-01-24
Support for large queries on the importer and add member pages. Show collection member publish status in listing.

brittnylapierre:
As Beth was testing she ran into a few problems (https://github.com/crkn-rcdr/Access-Platform/issues/464)

The requests to couch are chunked (for now, we can look into the couch configs to increase the timeout later) to handle very large arrays of slugs. (See changes to files in couch-utils) A check for duplicates was also added to the request for adding members, since Beth noticed she had duplicate members added.

There are some more checks done in the slug search and the input was moved to a re-usable component to remove code duplication between the importer and member addition. See /components/access-objects/PrefixSlugSearchbox.svelte and the handleSlugsChange method for the checks done.

I quickly added the publish status of collection members as I was touching up the member search with the PrefixSlugSearchbox component. Because during the import flow, it's not obvious that you need to publish your manifest after adding it to your collection. Having the publish status there reminds people to do so and lets them save a lot of time looking through each item to see if it's published or not.

### DIFF
--- a/packages/couch-utils/src/DatabaseHandler.ts
+++ b/packages/couch-utils/src/DatabaseHandler.ts
@@ -106,24 +106,6 @@ type BulkGetResponse<T extends Document> = {
 };
 
 /**
- * A helper method that breaks apart large arrays.
- * See: https://github.com/haio/chunk-array/blob/master/index.js
- */
-function chunkArray(array: any[], n: number) {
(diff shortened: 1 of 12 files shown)


SaschaAdler (approved):
I think you have a dangling `console.log` in the lapin changes. Otherwise, looks fine to me.

PR #471, opened 2022-01-24
Adding dmdType check in couchdb updateObject design doc  (forceUpdate & editObject) + fixed a bug with the create new collection/manifest button

brittnylapierre:
Two fixes, one for creating a new collection bug introduced when the updateInternalmeta status feature was added to the editor, the other checks for dmdType before updating updateInternalmeta to trigger hammer

### DIFF
--- a/services/couchdb/prelude.js
+++ b/services/couchdb/prelude.js
@@ -85,9 +85,10 @@ exports.updateObject = (doc, user) => {
     doc.staff = { by: user, date: now };
   }
 
-  if (doc.updateInternalmeta) delete doc.updateInternalmeta;
-  doc.updateInternalmeta = { requestDate: now };
-
+  if (doc.dmdType) {
+    if (doc.updateInternalmeta) delete doc.updateInternalmeta;
+    doc.updateInternalmeta = { requestDate: now };
+  }
   doc.updated = now;
 };
 
--- a/services/admin/src/lib/components/access-objects/EditorForm.svelte
+++ b/services/admin/src/lib/components/access-objects/EditorForm.svelte
@@ -218,7 +218,7 @@ This component displays the non content properties for an access editorObject an
       </span><br /-->
     </div>
     <div class="cache-status">
-      {#if chacheStatus.found && chacheStatus.result}
+      {#if chacheStatus?.found && chacheStatus.result}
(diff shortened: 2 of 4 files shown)


services/couchdb/prelude.js
```diff
@@ -85,9 +85,10 @@ exports.updateObject = (doc, user) => {
     doc.staff = { by: user, date: now };
   }
 
-  if (doc.updateInternalmeta) delete doc.updateInternalmeta;
-  doc.updateInternalmeta = { requestDate: now };
-
+  if (doc.dmdType) {
+    if (doc.updateInternalmeta) delete doc.updateInternalmeta;
+    doc.updateInternalmeta = { requestDate: now };
```
RussellMcOrmond, line 90: While this avoids the error, it does so silently.

I wonder if you should also do checks in https://github.com/crkn-rcdr/Access-Platform/tree/main/services/couchdb/access/design/access/updates so that the caller is sent back a warning.

IE: In https://github.com/crkn-rcdr/Access-Platform/blob/main/services/couchdb/access/design/access/updates/publish.js you could add:



```
  if (!doc.dmdType) {
    return errorReturn(`Descriptive Metadata missing`);
  }
```

RussellMcOrmond:
I think it would be confusing if the UI suggested that the Access document was published, but where it silently didn't attempt to generate the cache files (updateInternalmeta).   I think an error and not changing the document (IE: no change to doc.public , doc.updateInternalmeta or doc.updated) would be preferable.

brittnylapierre:
Sounds good, right now the UI doesn't let people press publish unless dmdType is set, but we could have times where a script is trying to publish something, or maybe some other tool in the future will need to. so, it makes sense to add

  if (!doc.dmdType) {
    return errorReturn(`Descriptive Metadata missing`);
  }

to the publish.js design doc.

I don't think I need to add it as an error return to editObject or any of the others? This is used to update slugs, labels, etc for anything that is published or unpublished. In that case we can just not trigger the doc.updateInternalmeta = { requestDate: now }; with the if (doc.dmdType) check?

RussellMcOrmond:
At the moment the only script that is trying to publish things without going through the UI is "makepublic", and your update solves that problem.

RussellMcOrmond (approved):

PR #474, opened 2022-01-25
Admin app housekeeping and small fixes

brittnylapierre:
1. Changed listFromKeys to a mutation so that querying for many AIPs on the dipstaging page does not bread
2. Add a call to trim() for slugs on the metadata updater page so that trailing/leading whitespace stops causing an error
3. Collections that have no members have the 'add members' search opened by default
4. Changed the colour scheme to be more minimal, eliminating many shades of grey
5. Fixed any imports that used '../' instead of '$lib/components'
6. Fix a big that meant the new prefix/search componenet did not update the slugs when the prefix changed
7. A quick bulk member publish/unpublish feature because this part of the workflow was super painful


brittnylapierre:
Just going to merge these improvements and fixes for Beth to test with me tomorrow and give feedback

PR #475, opened 2022-01-25
When launching new smelt, clear past OCR data processing.

RussellMcOrmond:
A trivial fix....

I believe that whenever a new Smelt is requested that any past OCR processing of that package should be cleared.  If the Smelt is successful, then the OCR processing will get re-run some time in the future.


brittnylapierre (approved):
Makes sense!

PR #484, opened 2022-02-04
Ocr batch form

brittnylapierre:


RussellMcOrmond:
Suggestion to have the database for this be "ocrtask" rather than "ocr" to better match what we did with "dmdtask".

RussellMcOrmond (approved):
I think we should merge this and allow John and Beth to review as well.

PR #485, opened 2022-02-16
Sub types for dmd tasks and test listing page

brittnylapierre:
I made changes to allow for parsing the different states a dmd task will be able to be in: 
1. Parsing
 2. Parsed (Reporting that background processing could split or validate the metadata file. The processor attempted to parse each individual item it extracted from the file; those items that parsed successfully have their own attachments added to the document.)
3. Updating
4. Update Success/warning/fail (Reporting that each item's metadata file was updated, or not.)

It's clearest in the fixtures I made what the format is for each.

I renamed ParseRecord to ItemProcessRecord.

I mostly used refine methods to differentiate between types.

You can see the tests & this listing to verify parsing for the types are working:
https://access-dev.canadiana.ca/dmd
Screenshot from 2022-02-16 17-38-38


RussellMcOrmond (approved):
Much easier to read.

Q: I'm wondering if we have decided to store in ISO date format rather than in seconds since?  Having one format rather than two possible seems to me to make the code easier to test as you don't need to test two different scenarios and potentially process dates differently in different code.

SaschaAdler (approved):
I don't have a lot of context for where this is at right now but I will say that the code looks good and extends the original functionality in a way that makes sense to me.

PR #498, opened 2022-02-23
Dmd listing UI

brittnylapierre:


RussellMcOrmond (commented):
There appears to be missing fixtures for when the parsing stage fails.

In this case `items` would be undefined, and `process.succeeded=false`

Comment on schema changes:

Parsed and Update *both* have Success/warning/fail possibilities.

RussellMcOrmond:
I went to https://access-dev.canadiana.ca/dmd/0f6851148dd5ed408ad7b7355f9cd9f4 , and kept choosing different prefixes. The tool seemed to keep adding prefixes.  You should probably remember the ID that was there at the start, and only store when "Load Metadata" button pressed.

I hadn't chosen "Load into New Access" or "Load for saving into preservation", so I'm not sure which database it was looking into.  A new lookup is needed when that setting is toggled.

In the future you may opt to allow the operator to chose the ID - it would be like "Import into Access", but with the box being reversed -- Green means it was found (in the chosen database) and can be used to load, red means it can't be checked to load.

When I choose "Preview Metadata", it says "Could not load preview. Please contact the platform team for assistance."

RussellMcOrmond:
The deployed version of `dmdtask` has: https://github.com/crkn-rcdr/cihm-metadatabus/blob/71cdd9829944276ceaae08ae19f8412f34d2e252/CIHM-Meta/lib/CIHM/Meta/Dmdtask.pm#L240-L245 where it is aware of items[] possibly existing in the future when the task is triggered.

Tthe deployed version should work perfectly fine with both the current and this new web interface (currently giving a warning it didn't do any storing).

Keeping backward compatibility with database changes makes coordinating the various updates easier. Backward compatibility for a few versions should be maintained unless it is absolutely necessary to break.

It should be possible to run the new web interface with the old `dmdtask` micro-service, as well as running the old web interface with the future `dmdtask` micro-service.

RussellMcOrmond:
* Load into New Access
* Load for Saving to Preservation

Language seems awkward to me.  Curious why "Load into Access" and "Load into Preservation" wasn't used?  There is no such thing as "load into old access", so there wouldn't be confusion.

Not sure what "Load into Saving to Preservation" means, as what it does is load into the OAIS packaging database.

RussellMcOrmond:
https://access-dev.canadiana.ca/dmd/f296cdd98ef39a425878357abadecb54

Set prefix to oocihm, clicked "load into New Access", checked first item only (oocihm.lac_reel_c1472), clicked "Load Metadata".

Database didn't change.  I expected "shouldStore" to be true.


```
  "items": [
    {
      "label": "C-1472",
      "parsed": true,
      "output": "issueinfo",
      "message": "",
      "id": "lac_reel_c1472",
      "shouldStore": false
    },
    {
      "shouldStore": false,
      "id": "lac_reel_c5622",
      "parsed": true,
      "label": "C-5622",
      "message": "",
      "output": "issueinfo"
    },
```

RussellMcOrmond:
Checked flipping between prefixes, and that is now working.

I went to https://access-dev.canadiana.ca/dmd/0f6851148dd5ed408ad7b7355f9cd9f4 and when I chose "oocihm" prefix, they were both green.  I flipped to "Load for Saving to Preservation" and it remained green, even though I removed the oocihm.8_00525_1 document from the `wipmeta` database.

RussellMcOrmond:
Generic schema design question: for the `ItemProcessRecord`, I am wondering why it was decided to have both a `destination` and a `shouldStore` field?

Is the intention to be able to set the destination, but that the field is ignored unless shouildStore is set to true?  That is how I am interpreting it, so checking if that is intended.

```
if (item.shouldStore == JSON::true && item.stored != JSON::true) {
    switch(item.destination) {
       case "access" { storeAccess(item);}
       case "preservation" { storePreservation(item); }
       else {
          warn "Destination Unknown! I left my job, my boss, my car and my home";
       }
    }
}
```

brittnylapierre:
> 

Hey Russell, that's the intention. I could change it so that the destination is set to null if it is unselected in the UI, I would just need to add a seperate state tracker for each item's checkbox input in the table.

Right now it looks like:
               <input
                  type="checkbox"
                  bind:checked={item.shouldStore}
                  on:change={checkIfAllItemsSelected}
                />

I would need to do something like
               <input
                  type="checkbox"
                  checked={item.destination ? true : false}
                  on:change={() => {
                      if false then set item.destination to null;
                      checkIfAllItemsSelected()
                  }
                />


Let me know if you want me to make this change & get rid of the shouldStore.

brittnylapierre:
Otherwise I think all of the above items are fixed. I ran this one: https://access-dev.canadiana.ca/dmd/746f167bc940060dee8742484c9d8cd7

Which successfully showed it loading. Then it got the stored not implemented result. The item's stored: boolean wasn't set, because that's not implemented yet, but when I manually do it in the database, it shows the correct success page.

I also ran it in the old front end after doing that without issue. It does show the stored not implemented yet warning though.
https://access.canadiana.ca/dmd/746f167bc940060dee8742484c9d8cd7

Let me know if anything else should be changed / if I should get rid of the shouldStore variable in the item object.

RussellMcOrmond:
Went to https://access-dev.canadiana.ca/dmd/0ec219d798d8828a3c555759616fe7e7

chose "oocihm" and "Load into OAIS Packaging Database".   The `destination` and `shouldStore` are as I would expect, but the `id` field is missing the prefix.

I've pushed what I did so far in https://github.com/crkn-rcdr/cihm-metadatabus/pull/27 to production in case you want to use that while testing as well.  It outputs a warning instead of actually doing something.

Logic of having both `destination` and `shouldStore` is fine.  I only wanted to confirm I understood the intention correctly.  It is possible when adopting Archivematica that we will want to create a new destination.

brittnylapierre:
Okay! I fixed it so the prefixed id is saved as well now.

RussellMcOrmond:
https://access-dev.canadiana.ca/dmd/f9ac77053d87b9be4384d10ab0813d1a

This metadata file has the same ID twice.  I'm curious for your thoughts on how to handle this: should the web interface detect this, or should we leave the error as it currently happens?  There is an update conflict with loading to preservation, but it just saves twice in the case of Access (the last one is the one kept).

RussellMcOrmond:
I tried uploading "McGillMapsMetadata-20220308.csv", and got an "Error: Metadata upload request failed."

```
Request URL: https://access-dev.canadiana.ca/api/dmdTask.create?batch=1
Request Method: POST
Status Code: 500 
Remote Address: 127.0.0.2:443
Referrer Policy: strict-origin-when-cross-origin
```

```
access-lapin-prod | missing json key: canProcessValidation
access-lapin-prod | 2022-03-24T13:13:06.839Z -- Error: dmdTask.create (mutation): INTERNAL_SERVER_ERROR
access-lapin-prod |  Unknown error
access-haproxy-prod | 172.30.0.1:44578 [24/Mar/2022:13:12:41.745] public~ access/local 11665/0/1/13433/25098 500 444 - - ---- 1/1/0/0/0 0/0 {https|access-dev.canadiana.ca} "POST https://access-dev.canadiana.ca/api/dmdTask.create?batch=1 HTTP/2.0"

```

RussellMcOrmond:
Tasks in "Awaiting Parsing" say "N/A N/A"?

https://access-dev.canadiana.ca/dmd/cff9b4f74416a0894771bc9ef8bd114d

"Something went wrong"  -- this was in the "Parsing" phase, but the microservice isn't running so it isn't current.  The parsing phase doesn't have a "I started" CouchDB call like storing does.

RussellMcOrmond:
Q:  What should be done with the tasks like http://jarlsberg.tor.c7a.ca:5984/_utils/index.html#database/dmdtask/cff9b4f74416a0894771bc9ef8bccf4a which have a successfully uploaded file, but where "process" wasn't initiated?  It isn't in the queue.  Should we just document staff to delete these?

brittnylapierre:
Fixed the issues above - 

That task shows up in the awaiting parsing section. If everything succeeds, parsing will automatically be triggered for any file uploaded, so they will move to the parsing section (which includes our parse queued and parsing state.) So I can actually rename that section to a word that best describes 'unable to trigger parse.' Right now we let them either delete them or open an issue letting us know the parsing didn't work:

The open issue button links to:
https://github.com/crkn-rcdr/Access-Platform/issues/new?title=DMD+Task+file+not+uploading+correctly&body=DMD+task+id:+cff9b4f74416a0894771bc9ef8bccf4a&labels[]=bug

brittnylapierre:
Hey @RussellMcOrmond let me know if you think I can merge this with main. I could use a new branch for enhancements and bug fixes this week. Or, I can build a new image off of this branch again to push with my above updates. Whatever you think is best!

RussellMcOrmond (approved):
I was about to ask a similar question about  https://github.com/crkn-rcdr/cihm-metadatabus/pull/27

PR #503, opened 2022-03-02
Importer enhancement & slug set to AIP id on page load

brittnylapierre:
Please see the following screenshots, and let me know if anything should be changed!

Importer, refresh button allows the same slug to be searched again.
importer

Editor, refresh button allows the same slug to be searched again.
editor

There will be a loading status on search (helpful for slow connections)
loading

I've tested the import process and slug editing process.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -42,7 +42,7 @@ services:
     container_name: access-noid
     depends_on:
       - haproxy
-    image: docker.c7a.ca/noid:20210226
+    image: d948bd40f1fc
     volumes:
       - /noid/dbs
     networks:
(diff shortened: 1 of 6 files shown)


RussellMcOrmond:
Is this ready to review?  I did `pnpm prod`

I went to: https://access-dev.canadiana.ca/smelter/find

Clicked on Sunday, which I knew had some LAC Reel's. Date: 2022-03-06

Text says "The slug entered is currently in use...." , except that it isn't used.


I then put in one I knew was used: oocihm.lac_reel_c6694

It gave the "slug in use".  When I hit the reload button without changing the slug, the "Slug in use" went away...

The "The slug entered ..." text remained in all cases.

RussellMcOrmond (approved):
I like the hover of the message -- red when in use, all ready to go when slug is available.

And the "Retry look-up" over the retry button.

PR #505, opened 2022-03-09
Updates for manifest PDF generating

SaschaAdler:
Sets up a queue and some update functions for the PDF generator.

I've also updated our fixture files to the current versions, to ensure that I can test off of canvases that actually exist. Apologies that that adds a bunch of noise to the PR.

The way I've set this up will allow for multiple PDFs per manifest, without overhauling the current multi-page PDF set-up.


RussellMcOrmond:
I'm confused by `rendering: z.array(LabelledFileRef).optional(),`

If the Swift Object ID is based on the NOID of the manifest and a file extension, how can there be an array of them?  Is the intention to include the array index or some additional ID within the Swift object ID?


Wondering if this is time to bring #231 into the repository, if there are any intended changes.

RussellMcOrmond:
Interesting idea to use a different design document rather than different names for the queue, update, etc.  We may want to think about that as a team whether we want to adopt that elsewhere. There are costs and benefits for each worth thinking about.

SaschaAdler:
> I'm confused by `rendering: z.array(LabelledFileRef).optional(),`
> 
> If the Swift Object ID is based on the NOID of the manifest and a file extension, how can there be an array of them? Is the intention to include the array index or some additional ID within the Swift object ID?
> 
> Wondering if this is time to bring #231 into the repository, if there are any intended changes.

I have a comment in the code proposing a way to do this, but I think I want to amend it to something like (e.g.)

```
access_files/69429/m0bz6154gh7n_pdf_001-126.pdf
```

Using `pdf` twice is less than ideal, but it means that we can do an object listing with the prefix `$noid_pdf` and get all of the pdf objects for the given noid, even if we do more "extensions" of object names like this. We could use `rendering` instead of `pdf` to avoid repetition if we really want.

We also don't have to specify the image range for each PDF, but it might be a nice thing to provide implicitly in the filename.

I'll add whatever we settle on to wherever the contents of #231 are going to end up. (Probably just an .md file in the root of this repo.)

> Interesting idea to use a different design document rather than different names for the queue, update, etc. We may want to think about that as a team whether we want to adopt that elsewhere. There are costs and benefits for each worth thinking about.

I went with this so that updating the views won't gum things up for already live functionality, but if we prefer to throw everything in the same design doc, we can.

RussellMcOrmond:
[Swift object storage](https://docs.openstack.org/api-ref/object-store/?expanded=get-object-content-and-metadata-detail#get-object-content-and-metadata) supports a `filename` get parameter, which is a good thing to remember.  We don't have to have the server suggest the same name during a download as how we stored it.  For instance, we could URL-encode the object name so that the part before the slash in the $noid is kept as part of the suggested filename.

In the case of metadata we used a slash as a delimiter , so we could search $noid/ to find objects related to that noid. The fact that we use $noid.$ext doesn't mean we can't also use $noid/ in the same container.

In code I've used split('/',$object) and then pop, knowing that while there won't be further slashes in the names that the noid currently has a slash (and future noids might not).

PR #513, opened 2022-04-05
Svelte kit update to allow for file uploads

brittnylapierre:
An update to svelte kit allows us to send files to the back-end the regular way.


brittnylapierre:
Hey @RussellMcOrmond A lot of the changes in here are really the changes in the filters branch + sveltekit upgrades.

I'm wondering if you would like me to use: services/admin/src/routes/dmd/[etag]/[id]/[fileName]/upload.js

To send a put request directly to couch directly from the admin tools svelte-kit back-end.

Or, would you like me to use the "upload" route in the rest interface? packages/hare-router/src/routes/dmdTask.ts

Let me know your suggestions. If I send the put request through the svelte-kit backend, I would just need to add a seperate request to create the task in the database to create the task first, then I can pass in it's id and rev and put the attachment to it.

The other method, through the rest interface, uses a function that creates the task and uploads the attachment.

Whichever method you think is best is the one I will take.

Thanks!

RussellMcOrmond:
Both options look good to me.  Are both sets of routes wrapped by the required authentication?  You want to ensure that only authorised staff are able to use either of these routes.

The first looks similar to what I was able to do in Sapindale, but I used logic similar to what you are doing with the REST request (IE: a PUT without the `[etag]/[id]/[fileName]` which would create a new document, upload the attachment, etc).

If neither is using the RPC, what is the major difference between the `server.put` with hare-router and `put` with the new svelte?  They both seem to give you the full body of the PUT request, and allow you to look at all the headers.

brittnylapierre:
I think the rest method is more secure/formal because it goes through the database handler that Sascha made, which takes advantage of nano.

The put fetch request from the sveltekit backend still needs people to log in but it uses a basic auth header for the fetch request.

I think if it looks good I'll keep it using the rest option (and remove the other file.) Because it will be good for consistency since all other features go through the api(s) to talk to couch.

If you want, I can build an image off of this for people to test the new upload. It might be useful to make a staging environment someday where the users could test alternate implementations without interrupting their work, now that the split is live.

RussellMcOrmond:
Was `ADMIN_URL_EXTERNAL` used anywhere in the code?  I'm not aware of any link that allows staff to go from the access admin interface to Sapindale...

RussellMcOrmond:
We will need to document what the HARE_URL environment variable means.  What you mentioned in the other issue might be a setting that worked for you in your dev environment, but won't be the setting in production.

Is this a URL that the web browser is intended to access?  If so, then there will be a change to the haproxy config as well. In a production environment the browser can't connect to anything that isn't going through the proxy.

Is HARE a separate web server, or is it intended to be a path within access.canadiana.ca ?

We'll chat more, and get the PR complete.

RussellMcOrmond:
As part of the ticket that is created to deploy the new Docker image to production:

* Document any new environment variables, and their meaning
* Document any new volume mounts for data that should persist.
* Document any changes to docker-compose, especially new containers (like "access-hare"), so the equivalents can be made in production.
* Document any new network connections that need to work.  For this PR there is a path between what in the dev environment is called "access-admin" and "access-hare".  If there was a new path from HAProxy, that would need to be documented as the HAProxy config would need to change as well.


Just saying "here is a new image" is interpreted to mean that none of the above changed from the previous image version.

brittnylapierre:
Awesome thanks Russell! I'm going to add your notes above to our Admin-Platform readme by the deploy section

PR #528, opened 2022-07-13
Changes to `canvases` database relating to Normalization of access files.

RussellMcOrmond:
This PR started out as a place to put some views to get statistics from the `canvases` database.

There is a schema change needed to handle the fact that the image file we store in Access Swift storage may no longer be a copy of the file that was pulled from Preservation storage.

In `/_design/access/_view/cihmsource` we were assuming that doc.source.path, doc.master.size and doc.master.md5 would uniquely confirm that a CIHM AIP pathname was to the same file (given SIP updates are allowed, paths are not unique identifiers).  Now doc.source and doc.master could refer to entirely different files, with different md5 and size.


RussellMcOrmond:
There is other cleanup worth looking into later, but it is worth reviewing and merging this sooner.

PR #534, opened 2022-08-24
Admin bug fixes - module versions, open in a new tab module error (lodash), and pdf publishing

brittnylapierre:
My dev env was failing to build so I set the svelte version to the last stable release explicitly 

Then I fixed the issue for the module error on new tab load

Then I fixed the issue of not being able to publish pdfs

### DIFF
--- a/services/admin/src/hooks.ts
+++ b/services/admin/src/hooks.ts
@@ -82,7 +82,7 @@ export async function handle({ event, resolve }) {
 	}
 
 	// I don't know why this is necessary; might have to do with https://github.com/sveltejs/kit/issues/2102
-	if (env.mode === 'production' && event.url.pathname.startsWith('/')) {
+	if (env.mode === 'production' && event.url.pathname.startsWith('/static')) {
 		try {
 			const filepath = pathJoin(process.cwd(), 'static', event.url.pathname.slice(8));
 			const file = readFileSync(filepath);
--- a/services/admin/src/hooks.ts
+++ b/services/admin/src/hooks.ts
@@ -82,7 +82,7 @@ export async function handle({ event, resolve }) {
 	}
 
 	// I don't know why this is necessary; might have to do with https://github.com/sveltejs/kit/issues/2102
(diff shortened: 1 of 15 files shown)


RussellMcOrmond (approved):
I didn't build/test, but from a scan of the code it all makes sense.

Interesting that the editor has different ideas about formatting -- did a VS Code module change?

services/admin/src/routes/index.svelte
```diff
@@ -64,7 +64,7 @@
 <!--div class="title">
     <img
       class="logo"
-      src="/static/canadiana-pa-tag-color.png"
+      src="/canadiana-pa-tag-color.png"
```
RussellMcOrmond, line 67: I find this interesting -- a change in where the image is located?

services/admin/src/routes/__layout.svelte
```diff
@@ -134,7 +130,7 @@
 <style>
 	@font-face {
 		font-family: 'Roboto';
-		src: url('/static/fonts/Roboto-Regular.ttf') format('truetype');
+		src: url('/fonts/Roboto-Regular.ttf') format('truetype');
```
RussellMcOrmond, line 133: Same here and other places.   Did /static go away?

Are .svelte files generated by a build process, or are these source files?  I thought source files, but maybe I'm wrong?

services/admin/src/hooks.ts
```diff
@@ -82,7 +82,7 @@ export async function handle({ event, resolve }) {
 	}
 
 	// I don't know why this is necessary; might have to do with https://github.com/sveltejs/kit/issues/2102
-	if (env.mode === 'production' && event.url.pathname.startsWith('/static/')) {
+	if (env.mode === 'production' && event.url.pathname.startsWith('/static')) {
```
RussellMcOrmond, line 85: If /static is no longer used, is this needed?  Was /static about a specific technique that is no longer used?

brittnylapierre: Removed it! Yeah it started to complain about the assets having /static so I think they removed the need for this

PR #538, opened 2022-09-25
Update cache for unordered collections - add and remove members routes

brittnylapierre:
Tested all of the add/remove operations all working as expected on my end, triggering updates for only the members that were added or removed for unordered collections, with multi-part functionality remaining the same as it was

### DIFF
--- a/packages/lapin-router/src/routes/collection.ts
+++ b/packages/lapin-router/src/routes/collection.ts
@@ -239,7 +239,11 @@ export const collectionRouter = createRouter()
         }
 
         // Don't hold up the response but force update to these new members
-        if ("behavior" in collection && collection.behavior === "multi-part") {
+        if (
+          "behavior" in collection &&
+          (collection.behavior === "multi-part" ||
+            collection.behavior === "unordered")
+        ) {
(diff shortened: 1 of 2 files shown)


RussellMcOrmond (approved):
I am guessing with all the `console.log` , that you have done some testing and feel comfortable?

I haven't followed the logic closely to know what `filteredMembers` or `memberIds` are, but I expect they are changed members.

Note: At the moment the only `collection.behavior` values we support are "multi-part" and "unordered".  You can decide if you want to check such things explicitly now prior to there being a third option.  You may even reverse the check and have it alert you if a different value shows up.

brittnylapierre:
> I am guessing with all the `console.log` , that you have done some testing and feel comfortable?
> 
> I haven't followed the logic closely to know what `filteredMembers` or `memberIds` are, but I expect they are changed members.
> 
> Note: At the moment the only `collection.behavior` values we support are "multi-part" and "unordered". You can decide if you want to check such things explicitly now prior to there being a third option. You may even reverse the check and have it alert you if a different value shows up.

Yes I do! I tested all of the scenarios this morning. I'll make an image and post so everyone knows to let us know if the problem resolves. I am getting feedback from the staff right now for the trigger button issue

PR #574, opened 2023-01-25
Adding an interface for interacting with updateInternalmeta and hammer queues

brittnylapierre:
- [ ] having the slugs in that list having a link to the specific item would be great.  I was about to cut-and-paste "oocihm.lac_reel_c1870" to then go to https://access.canadiana.ca/object/edit/69429/m05t3fx78707
- [ ] You likely want to transform line breaks to, or put the text into some sort of block that will display the newlines. (an input, pre, etc).


RussellMcOrmond:
General suggestion for messages coming from thee back-end tools.  The text has line breaks, which aren't shown if you just output the text into the HTML.

You likely want to transform line breaks to <br>, or put the text into some sort of block that will display the newlines. (an <input>, <pre>,  etc).

Issue #90, opened 2020-04-22
Design schemas and plan for tools to generate Manifest-level multi-page PDFs


RussellMcOrmond:
https://github.com/crkn-rcdr/Access-Platform/blob/master/Databases/schemas/manifest.json has an "updatePdf" field that can be used to create a view for a queue.

* For every manifest where this field is blank, or where the updatePdf.requestDate>updatePdf.processDate
* Check if there is an ocrPdf file in the manifest, but no ocrPdf in the canvases.
  * split the multi-page ocrPdf to generate separate single-page ocrPdf files, and return.
* Check if there is an ocrPdf in all the canvases, and the count or size is not higher than a maximum (Policy needed)
  * join the single-page canvas ocrPdf files into a multi-page ocrPdf, and store associated with manifest.

SaschaAdler:
It was only deprecated because the schema only contains the stuff required to get the split done. :)

RussellMcOrmond:
Note that #153  completed the first part of the work: "split the multi-page ocrPdf to generate separate single-page ocrPdf files."

The remaining work is to: "join the single-page canvas ocrPdf files into a multi-page ocrPdf, and store associated with manifest."

brittnylapierre:
Generating PDFs: 

- [x] always when OCR is made, 
- [ ] always when something is published, 
- [ ] and always when the canvases change order.

PR #576, opened 2023-02-03
CouchDB design for simple service to generate multi-page PDFs.

RussellMcOrmond:
https://github.com/crkn-rcdr/Access-Platform/issues/90

### DIFF
--- a/packages/data/src/access/Manifest.ts
+++ b/packages/data/src/access/Manifest.ts
@@ -6,6 +6,7 @@ import {
   ObjectListShort,
   ObjectListHandler,
 } from "./util/ObjectList.js";
+import { ProcessUpdate } from "../util/index.js";
 
 /**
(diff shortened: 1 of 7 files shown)


RussellMcOrmond:
@brittnylapierre ,

I'd like to propose adding an optional array to createOCRPDF  (via a new merge of ProcessUpdate ) that would be an optional array of canvas NOIDS which didn't have OCR data.

The current messages are mostly "Missing odfPdf for canvas index=0" which is not very useful.

* I think noids are more useful than indexes -- while a noid can be turned into an index for visualising, being able to copy that array into an https://access.canadiana.ca/ocr batch would be very useful.
* I think the simple message "Caught: Not all canvases have OCR data" is all staff would need to care about, with the details of which canvases being elsewhere.

If you agree, let me know if you want to make the changes to this PR or if you wish me to.

Issue #585, opened 2023-02-16
UI to Browse Image Canvases? Make manifest creation for "Import into Access" optional in UI.


RussellMcOrmond:
I will need part of this work to be done as part of cleaning up Numeris images, so I wanted to document the larger idea.

* New _view in https://github.com/crkn-rcdr/Access-Platform/tree/main/services/couchdb/canvas/design/access/views
  * Maybe add to "access" design document, or create a new "browse" design document 
  * View will be divided by the ["source" field](https://github.com/crkn-rcdr/Access-Platform/blob/main/packages/data/src/canvas/Canvas.ts)
  * First element of array uses as key is "from", which currently must be "cihm" (other "from" to be handled later)
    * Create string from "path" that removes "/data/sip/data/files/" and any extension (.tif, .jpg, .jp2)
    * Split string into array separating at ".", "/" or "_"
  * This view can be used to browse images without needing to use manifests, and can later be built into staff tools and a source for the IIIF manifest editor
* Make creating a manifest optional for  https://access.canadiana.ca/smelter (Possibly a checkbox that will clear the slug field in the database rather than storing an unused slug)
  * Update Smelter to not create a manifest if the slug field is undefined, only import images into Access files and update the `canvases` CouchDB database 

The staff tools enhancements will be work for @brittnylapierre in the future, and the microservices and new _view is work I will do myself.

See also: https://github.com/crkn-rcdr/cihm-metadatabus/issues/70 , which is also part of the post-PASplit clean-up.

RussellMcOrmond:
There are many questions about the UI for @brittnylapierre to decide (with input from Heritage Services).

* Browsing Image Canvases based on source.
  * Use only as part of Digitrati collaboration?
  * Add to current interface?
  * Only provide an API for a IIIF page viewer?
  * Does https://access.canadiana.ca/object/new/manifest use this, to bulk-add images to a newly created manifest?
  * Bulk options for creating many manifests from many sets of images at the same time?  Make equivalent interface to "Import into Access" for generating manifests from images already imported?

* Making Manifest creation for "Import into Access" optional
  * Default to not create?
  * Above interface used instead? Both, or new replaces old?

RussellMcOrmond:
To see how this works:

Go to  https://admin.canadiana.ca/ to confirm you are logged in. It will mess up the URL if you forget to do this step.


[...]

PR #593, opened 2023-02-27
 View to enable Browsing Image Canvases

RussellMcOrmond:
https://github.com/crkn-rcdr/Access-Platform/issues/585


RussellMcOrmond:
We could decide to merge these changes to _views early, and treat the changes to the UI as a separate PR.

PR #603, opened 2023-03-30
Bug fixes m27

brittnylapierre:


services/.DS_Store
RussellMcOrmond, line 1: @brittnylapierre ,

I suspect this was added by accident.  If you are using MacOS, then you'll want to include these files in the exclusion lists.

PR #608, opened 2023-04-18
Bulk labels

brittnylapierre:
Here is how I think the procedure should go, let me know if you have any thoughts or would like to change it:

1. Upload a CSV file containing 'slug' (full) and 'label' columns
2. A IIIF task is queued
3. The back-end scripts creates a blank items array on the task, opens the csv, iterate through the rows, attempts to set the value, and appends an object to the items array with the following info { slug, succeeded, message }. The script also sets the progress field as it iterates over the rows. When it is done, or if it errors, it sets updates the processStatus.
4. When the processStatus indicates the script is done working on the task, the task will show up in the completed area on the ui. 
5. Staff can download a CSV summarising the results of the task and giving status messages for succeeded and failed items (CSV is familiar to staff, and is less work to provide and edit than a UI showing the table of results. They also enable search and advanced filtering.)

Example result CSV sheet 1:
Screenshot 2023-04-18 at 8 32 13 AM

Example result CSV sheet 2:
Screenshot 2023-04-18 at 8 32 30 AM

6. Staff can delete any items under the completed expansion panel, too.

IIIF Task Structure:
/Access-Platform/packages/data/src/iiif/Task.ts

IIIF Task Couch Setup:
/Access-Platform/services/couchdb/iiiftask


brittnylapierre:
Hmm I didn't think about it yet but after thinking about it, I think we could do:

items : [
    { 
         slug: "bl.test",
         results: { // one sub-obj per column
                label : {
                       succeeded: true
                       message: ""
                },
                ... 
          }
    },
    ...
]

Or something like that? Not sure if you prefer arrays or hashes when it comes to these things.

